PR #13, opened 2021-10-18
Add scripts for PR metrics from github API

mpg:
These scripts allow to track a number of metrics about PRs over time, using the github API.

So far they've only be used by me on my machine, but they should be a shared team resource, so here's a PR. The current versions are not intended to be final, they're a starting point that we can evolve in future PRs as needed.

(Some previous review history at https://github.com/ARMmbed/mbedtls-test-old/pull/153 (private).)


gilles-peskine-arm (changes requested):
I checked that I could generally understand what the scripts are doing and that it matches what they're documented to do. I also checked that I could run the scripts on my machine and that their output looks plausible.

I left a number of comments suggesting improvements to the documentation, the coding style or the code structure. The only one I consider to be a blockers is about the script docstrings not really explaining how to run the scripts. I would prefer the other documentation improvements as well. Non-comment changes are merely nice to haves or things to do in the future, but not obstacles to merging what's already there. In the future, I would like to [pass the same checks on Python code as in the mbedtls repository](https://github.com/ARMmbed/mbedtls-test/issues/17).

gilles-peskine-arm:
Labeling "size:S" based on review time (I would guess a size:M based on coding time, but this is relatively easy to review since checking what a library function does is easier than looking for a library function that does a given thing).

pr-metrics/pr-lifetime.py
```diff
@@ -29,7 +29,12 @@
     else:
         lt = (end - beg).days
 
-    q = quarter(beg)
+    # Shit one month (that is, for q2 count March to May, not April to July).
+    # This is because we want to measure this at the end of each quarter, but
+    # including PRs raised too recently skew the results. Shifting one month
+    # means we had time to look at the PR by the time we generate quaterly
+    # metrics.
+    q = quarter(beg - timedelta(days=30))
```
tom-cosgrove-arm, line 37: Is this going to invalidate previous sets of statistics?

mpg: It is. Which indeed might be a problem.

Perhaps a better approach is to decide that we just can't report the numbers for Q4 at the beginning of Q1, and only report numbers one quarter later.

FWIW, even with this 30-days shift, we run into problems trying to generate our Q4 statistics today, because less than half of the community PRs created until the 1st of December have been closed, so we can't compute (even an estimate of) the median of the time it took to close them.

Basically we can't compute even a temporary value of the median until we've merged half of the PRs created in that period, and after that, the value returned by the script might be an overestimate of the actual median until time elapsed since the end of the quarter becomes larger than the value returned by the script.

The more I think about it, the more I think the honest thing to do is wait until we can actually compute a value before we report it.

mpg: Reverted this commit and instead pushed something that finally handles the uncertainty properly (I hope!): by reporting it. More precisely, excluding uncertain quarters from the graph, where I'm not sure how to report it, and reporting a lower bound (and upper bound when available) in the csv.

I checked the new path doesn't change results for past quarters.

pr-metrics/pr-backlog.py
```diff
@@ -0,0 +1,72 @@
...
+
+from prs import pr_dates, first, last, quarter
+
+from datetime import datetime, timedelta
+from collections import Counter
+from itertools import chain
+
+import matplotlib.pyplot as plt
+
+new_days = 90
```
daverodgman, line 14: There was a bit of a push to standardise ages for what is considered recent/old in OSS a while back. The thresholds picked were 15 and 90.

It might be useful to have a few extra ranges, e.g. <15, 15-90, 90-365, >365 to align better with this? If I have time I'll push an update.

daverodgman: Actually, thinking more, this is a lot like the median lifetime graph, but with a couple of thresholds which are based on age rather than percentiles. Would it be better to have a graph that shows e.g., median, 75th percentile, 95th percentile? Pros and cons either way I think, maybe worth exploring it if it's quick/easy to do but I'm not sure if it would be better or not?

mpg: Oh, I didn't know about the push for standardised thresholds, indeed it would be good to align with that (and add some of our own if needed). Unfortunately the way the script is structured currently makes it a very manual change (you can't just give a list of threesholds at the top and have everything else work automagically).

mpg: Regarding adding other percentiles to the median one, that's something I've been considering for a while, but the problem is over what set. Currently it's over "PRs created this quarter", which means, considering Q1 for the sake of concreteness we can only compute the median (or an upper bound for it) after we've closed at least 50% of PRs created in Q1. Fortunately, that's usually the case at the very beginning of Q2 when we prepare our report. (Even then, we might get only a range, as the value could still get lower if we closed a lot of PRs created near the end of Q1 at the very beginning of Q2.)

But for the 4th quartile, resp. 95th percentile, we'd need to have closed 75%, resp. 95% of PRs created in Q1 by the time we produce our report in early Q2, which realistically is not going to happen most of the time.

We could avoid the problem with incomplete data entirely by considering instead the set of PRs we _closed_ this quarter - there all the lifetimes are known for sure and we can do stats without uncertainty. But I doesn't really tell the same thing: for example, doing at lot of historical review one quarter would raise the median age of PRs closed this quarter, but that's still a good thing. So the data might become more difficult to interpret.

So, I think there are basically three ways to select PRs over which we make stats / grouping / etc for one quarter:
- PRs created this quarter;
- PRs still open at the end of the quarter (or any specific date);
- PRs closed this quarter;
and each set will give slightly different information.

Wdyt?

daverodgman: Agree, there isn't an easy answer or obvious better way. So let's leave as-is for now.

PR #33, opened 2022-01-28
Adding coverity push to nightly job

paul-elliott-arm:
## Description

Add a job to push to the coverity scan branch to the nightly tests.

Coverity push is conditional on new environment variable PUSH_COVERITY, so future nightly jobs will have to have that variable turned on in order to enable.

Coverity push will only happen if the nightly is running on development branch.

This closes https://github.com/ARMmbed/mbedtls-test/issues/14

## Testing.

This has been tested via the TEST_BRANCH variable, and found to be working : https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/361/

### DIFF
--- a/vars/gen_jobs.groovy
+++ b/vars/gen_jobs.groovy
@@ -575,6 +575,34 @@ mbedhtrun -m ${platform} ${tag_filter} \
     return jobs
 }
 
+def gen_coverity_push_jobs() {
+    def jobs = [:]
+    def job_name = "coverity-push"
+
+    if (env.MBED_TLS_BRANCH == "development") {
+        jobs[] = {
+            node('container-host') {
+                try {
+                    dir("src") {
+                        deleteDir()
+                        checkout_repo.checkout_repo()
+                        sshagent([env.GIT_CREDENTIALS_ID]) {
+                            sh 'git push -u origin HEAD:coverity_scan'
+                        }
+                    }
+                } catch (err) {
+                    failed_builds[job_name]= true
+                    throw (err)
+                } finally {
+                    deleteDir()
+                }
+            }
+        }
+    }
+
+    return jobs
+}
+
 def gen_release_jobs(label_prefix='', run_examples=true) {
     def jobs = [:]
 
@@ -609,6 +637,10 @@ def gen_release_jobs(label_prefix='', run_examples=true) {
 
     if (run_examples) {
         jobs = jobs + gen_all_example_jobs()
+    }
+
+    if (env.PUSH_COVERITY == "true") {
+        jobs = jobs + gen_coverity_push_jobs()
     }
 
     return jobs
--- a/vars/gen_jobs.groovy
+++ b/vars/gen_jobs.groovy
@@ -587,7 +587,7 @@ def gen_coverity_push_jobs() {
                         deleteDir()
                         checkout_repo.checkout_repo()
                         sshagent([env.GIT_CREDENTIALS_ID]) {
-                            sh 'git push -u origin HEAD:coverity_scan'
+                            sh 'git push origin HEAD:coverity_scan'
                         }
                     }
                 } catch (err) {


vars/gen_jobs.groovy
```diff
@@ -575,6 +575,34 @@ mbedhtrun -m ${platform} ${tag_filter} \
...
+
+    if (env.MBED_TLS_BRANCH == "development") {
+        jobs[] = {
+            node('container-host') {
+                try {
+                    dir("src") {
+                        deleteDir()
+                        checkout_repo.checkout_repo()
+                        sshagent([env.GIT_CREDENTIALS_ID]) {
+                            sh 'git push -u origin HEAD:coverity_scan'
```
yanesca, line 590: Why do we need the `-u` switch here?

vars/gen_jobs.groovy
```diff
@@ -611,6 +639,10 @@ def gen_release_jobs(label_prefix='', run_examples=true) {
         jobs = jobs + gen_all_example_jobs()
     }
 
+    if (env.PUSH_COVERITY == "true") {
```
yanesca, line 642: I think it would be better to add this as a parameter to the `gen_release_jobs()` method, like we do it with `run_examples`. At least for now; if it is an environment variable we would need to change the job configuration, to which we don't have access yet and would mean a communication overhead.

yanesca (commented):
@paul-elliott-arm could you please post a link to the successful run you mentioned above?

vars/gen_jobs.groovy
```diff
@@ -575,6 +575,34 @@ mbedhtrun -m ${platform} ${tag_filter} \
...
+                } finally {
+                    deleteDir()
+                }
+            }
+        }
+    }
+
+    return jobs
+}
+
```
gilles-peskine-arm, line 605: I guess the idea is to turn on `PUSH_COVERITY` in the nightly? But shouldn't there also be a way to do the pushing without invoking the test-related stuff (sanity checks, docker images, final report, etc.)?

vars/gen_jobs.groovy
yanesca: I don't have a problem with having it as an environment variable in the long term. In the short term however if we make it an environment variable, then we'll need to ask the OpenCI team that they add this parameter to the nightly on their side as well.

vars/gen_jobs.groovy
yanesca: I think for now it is fine to push it only in the nightly. If we want to do coverity without the nightly, we can just push directly. When we remove Travis from the equation, we might need to think about triggering Coverity without the test-related stuff, but until then I think this should be enough.

vars/gen_jobs.groovy
paul-elliott-arm: A slight muscle memory issue of mine. I conditioned myself to use -u, despite not really needing to.

vars/gen_jobs.groovy
paul-elliott-arm: Ok, reverted to just a boolean for now, we can as you say revisit this when the CI transition is over.

vars/gen_jobs.groovy
paul-elliott-arm: See above. We can for the minute still push to coverity_scan manually. If we remove travis, then as you say, an environment variable would be required in order to trigger scan manually. This can be looked at as and when any transition happens.

paul-elliott-arm:
Successful run of new code here : https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/361/

vars/gen_jobs.groovy
gilles-peskine-arm: By making this unconditional, we can't do a test run of the release code flow (e.g. to test changes to the nightly job) without pushing to coverity. I don't see how to resolve this without an environment variable. Controlling the Coverity push from the job configuration or parameters is a good thing.

Adding a parameter to the nightly is a trivial request. I don't understand why this is a concern at all.

vars/gen_jobs.groovy
```diff
@@ -611,6 +639,10 @@ def gen_release_jobs(label_prefix='', run_examples=true) {
         jobs = jobs + gen_all_example_jobs()
     }
 
+    if (push_coverity) {
```
gilles-peskine-arm, line 642: `push_coverity` should be an environment variable, set by the job configuration or parameters. See https://github.com/ARMmbed/mbedtls-test/pull/33#discussion_r801986371

vars/mbedtls.groovy
```diff
@@ -21,7 +21,7 @@ def run_tls_tests(label_prefix='') {
     try {
         def jobs = [:]
 
-        jobs = jobs + gen_jobs.gen_release_jobs(label_prefix, false)
+        jobs = jobs + gen_jobs.gen_release_jobs(label_prefix, false, false)
```
gilles-peskine-arm, line 24: Sorry for not catching this earlier, but I'm not convinced doing this in parallel with the testing is right. Shouldn't we only push to coverity if the functional tests have passed?

gilles-peskine-arm (changes requested):
Looks good to me apart from the push_coverity being forced on.

vars/gen_jobs.groovy
yanesca: It is indeed a trivial request, avoiding it still would have saved latency and effort. I want to get this done ASAP and not waste effort on it. At this point however, I feel that it is indeed more efficient to introduce that environment variable than continue discussing this.

@paul-elliott-arm I am sorry for flip-flopping on this.

vars/mbedtls.groovy
yanesca: Since everything landing on the release branches passes CI, the nightly should pass consistently. Practical experience supports this as well: we have mostly intermittent faults and infrastructure issues making the nightly fail and even in the face of this it passes consistently most of the time.

Even if the nightly fails, I don't see any harm in pushing to coverity anyway.

yanesca (dismissed):
Looks good to me and as mentioned above (https://github.com/ARMmbed/mbedtls-test/pull/33#discussion_r802406596), I am happy with re-introducing the environment variable as well.

paul-elliott-arm:
Put the environment variable back (apologies for force push). See comment above about when to push above.

Succesful test run : https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/365/

yanesca (approved):
Looks good to me.

gilles-peskine-arm (approved):

gilles-peskine-arm:
I added the `PUSH_COVERITY` parameter to the [nightly job](https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-nightly-tests/) (default on) and to the [manual release job](https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-new) (default off). Paul had already added it to the [testing job](https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing) (also default off).

So starting tonight we should have daily pushes to the coverity branch.

PR #31, opened 2022-01-14
Update and cleanup docker files

minosgalanakis:
This PR is implementing the following:

### Changes

* Align 16.04 to 18.04 Docker images
* Remove 16.04 related workarounds when possible
* Updated the ARMLMD_LICENSE_FILE files and removed workarounds in groovy scripts
* Created a 20.04 Docker Image file which is fully aligned to 18.04
* Added the `--security-opt seccomp=unconfined` to the docker run parameter
* Switched to distribution provided versions for abi-dumper, abi-compliance-checker libasn1

### Testing

Image are tested locally using the build and run scripts, and isolated builds

- [x] Test job mbedtls-release-ci-testing [347 ](https://jenkins-mbedtls.oss.arm.com/blue/organizations/jenkins/mbedtls-release-ci-testing/detail/mbedtls-release-ci-testing/347/pipeline)(RUN_BASIC_BUILD_TEST @ 16.04) -> Passed
- [x] Test job mbedtls-release-ci-testing  [347 ](https://jenkins-mbedtls.oss.arm.com/blue/organizations/jenkins/mbedtls-release-ci-testing/detail/mbedtls-release-ci-testing/347/pipeline)(RUN_BASIC_BUILD_TEST @ 18.04) -> Passed
- [x] Test job mbedtls-release-ci-testing  [347 ](https://jenkins-mbedtls.oss.arm.com/blue/organizations/jenkins/mbedtls-release-ci-testing/detail/mbedtls-release-ci-testing/347/pipeline)(RUN_BASIC_BUILD_TEST @ 20.04) -> Passed
- [x] Test job mbedtls-release-ci-testing [348 ](https://jenkins-mbedtls.oss.arm.com/blue/organizations/jenkins/mbedtls-release-ci-testing/detail/mbedtls-release-ci-testing/348/pipeline)(RUN_ALL_SH@ 16.04) -> Passed
- [x ] Test job mbedtls-release-ci-testing [348 ](https://jenkins-mbedtls.oss.arm.com/blue/organizations/jenkins/mbedtls-release-ci-testing/detail/mbedtls-release-ci-testing/348/pipeline)(RUN_ALL_SH@ 20.04) -> Passed
- [x] Testing RUN_ABI_TEST in mbed-tls-pr-test-parametrized [225 ](https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-test-parametrized/225/)
[...]

PR #34, opened 2022-01-31
Condition Open CI changes based on Jenkins URL

arthur-she:
Modify the build script to fit two CIs, Trustedfirmware OpenCI and ARM internal CI.

### DIFF
--- a/resources/docker_files/ubuntu-16.04/Dockerfile
+++ b/resources/docker_files/ubuntu-16.04/Dockerfile
@@ -121,7 +121,8 @@ RUN wget -q https://developer.arm.com/-/media/Files/downloads/compiler/DS500-PA-
 
 ENV ARMC5_BIN_DIR=/usr/local/ARM_Compiler_5.06u3/bin/
 ENV PATH=$PATH:/usr/local/ARM_Compiler_5.06u3/bin
-ENV ARMLMD_LICENSE_FILE=7010@10.6.26.52:7010@10.6.26.53:7010@10.6.26.54:7010@10.6.26.56
+ARG ARMLMD_LICENSE_FILE=7010@10.6.26.52:7010@10.6.26.53:7010@10.6.26.54:7010@10.6.26.56
+ENV ARMLMD_LICENSE_FILE=${ARMLMD_LICENSE_FILE}
 
 # Install ARM Compiler 6.6
 RUN mkdir temp && cd temp && \
--- a/vars/common.groovy
+++ b/vars/common.groovy
@@ -270,7 +270,7 @@ Logs: ${env.BUILD_URL}
 
 Failures: ${failures}
 """
-        subject = "${name} failed!"
+        subject = is_open_ci_env ? "TF Open CI ${name} failed!" : "Internal CI ${name} failed!"
         recipients = env.TEST_FAIL_EMAIL_ADDRESS
     } else {
         emailbody = """
@@ -278,7 +278,7 @@ ${coverage_details['coverage']}
 
 Logs: ${env.BUILD_URL}
 """
-        subject = "${name} passed!"
+        subject = is_open_ci_env ? "TF Open CI ${name} passed!" : "Internal CI ${name} passed!"
         recipients = env.TEST_PASS_EMAIL_ADDRESS
     }
     echo subject
--- a/vars/common.groovy
+++ b/vars/common.groovy
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2019-2021, Arm Limited, All Rights Reserved
+ *  Copyright (c) 2019-2022, Arm Limited, All Rights Reserved
  *  SPDX-License-Identifier: Apache-2.0
  *
  *  Licensed under the Apache License, Version 2.0 (the "License"); you may
@@ -41,7 +41,7 @@ import groovy.transform.Field
 @Field docker_ecr = "666618195821.dkr.ecr.eu-west-1.amazonaws.com"
 @Field docker_repo = "$docker_ecr/$docker_repo_name"
 
-@Field linux_platforms = ["ubuntu-16.04", "ubuntu-18.04"]
+@Field linux_platforms = ["ubuntu-16.04", "ubuntu-18.04", "ubuntu-20.04"]
 @Field bsd_platforms = ["freebsd"]
 @Field bsd_compilers = ["clang"]
 @Field all_compilers = ['gcc', 'clang']
--- a/vars/common.groovy
+++ b/vars/common.groovy
@@ -21,6 +21,9 @@ import java.security.MessageDigest
 
 import groovy.transform.Field
 
+/* Indicates if CI is running on Open CI (hosted on https://ci.trustedfirmware.org/) */
+@Field is_open_ci_env = env.JENKINS_URL ==~ /\S+(trustedfirmware)\S+/
+
 /*
  * This controls the timeout each job has. It does not count the time spent in
  * waiting queues and setting up the environment.
@@ -37,8 +40,8 @@ import groovy.transform.Field
     'cc' : 'cc'
 ]
 
-@Field docker_repo_name = 'jenkins-mbedtls'
-@Field docker_ecr = "666618195821.dkr.ecr.eu-west-1.amazonaws.com"
+@Field docker_repo_name = is_open_ci_env ? 'ci-amd64-mbed-tls-ubuntu' : 'jenkins-mbedtls'
+@Field docker_ecr = is_open_ci_env ? "trustedfirmware" : "666618195821.dkr.ecr.eu-west-1.amazonaws.com"
 @Field docker_repo = "$docker_ecr/$docker_repo_name"
 
 @Field linux_platforms = ["ubuntu-16.04", "ubuntu-18.04", "ubuntu-20.04"]
@@ -109,7 +112,12 @@ def get_docker_image(platform) {
     def docker_image = get_docker_tag(platform)
     for (int attempt = 1; attempt <= 3; attempt++) {
         try {
-            sh """\
+            if (is_open_ci_env)
+                sh """\
+docker pull $docker_repo:$docker_image
+"""
+            else
+                sh """\
 aws ecr get-login-password | docker login --username AWS --password-stdin $docker_ecr
 docker pull $docker_repo:$docker_image
 """
--- a/vars/common.groovy
+++ b/vars/common.groovy
@@ -238,7 +238,8 @@ def maybe_notify_github(context, state, description) {
         description = description.take(MAX_DESCRIPTION_LENGTH - 1) + '…'
     }
 
-    githubNotify context: "${env.BRANCH_NAME} ${context}",
+    content = is_open_ci_env ? "TF OpenCI: ${env.BRANCH_NAME} ${context}" : "Internal CI: ${env.BRANCH_NAME} ${context}"
+    githubNotify context: content,
                  status: state,
                  description: description
 }
--- a/vars/gen_jobs.groovy
+++ b/vars/gen_jobs.groovy
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2019-2021, Arm Limited, All Rights Reserved
+ *  Copyright (c) 2019-2022, Arm Limited, All Rights Reserved
  *  SPDX-License-Identifier: Apache-2.0
  *
  *  Licensed under the Apache License, Version 2.0 (the "License"); you may
@@ -220,16 +220,10 @@ scripts/min_requirements.py --user
                 }
                 dir('src') {
                     checkout_repo.checkout_repo()
-                    /* ARMLMD_LICENSE_FILE is supposed to be set in the
-                     * Dockerfile, but the value there is out-of-date and we
-                     * have trouble re-building the images (due to the death
-                     * of Python 2), so we override it here as a temporary
-                     * work-around. */
                     writeFile file: 'steps.sh', text: """\
 #!/bin/sh
 set -eux
 ulimit -f 20971520
-export ARMLMD_LICENSE_FILE="7010@10.6.26.52:7010@10.6.26.53:7010@10.6.26.54:7010@10.6.26.56"
 export MBEDTLS_TEST_OUTCOME_FILE='${job_name}-outcome.csv'
 ${extra_setup_code}
 ./tests/scripts/all.sh --seed 4 --keep-going $component
@@ -528,12 +522,9 @@ pip install -r requirements.txt
                             if (example == 'atecc608a') {
                                 tag_filter = "--tag-filters HAS_CRYPTOKIT"
                             }
-                            /* See gen_all_sh_jobs() regarding setting of
-                             * ARMLMD_LICENSE_FILE here. */
                             sh """\
 ulimit -f 20971520
 . $WORKSPACE/mbed-venv/bin/activate
-export ARMLMD_LICENSE_FILE="7010@10.6.26.52:7010@10.6.26.53:7010@10.6.26.54:7010@10.6.26.56"
 mbed compile -m ${platform} -t ${compiler}
 """
                             for (int attempt = 1; attempt <= 3; attempt++) {
--- a/vars/gen_jobs.groovy
+++ b/vars/gen_jobs.groovy
@@ -25,6 +25,8 @@ import groovy.transform.Field
 //Record coverage details for reporting
 @Field coverage_details = ['coverage': 'Code coverage job did not run']
 
+@Field armlmd_license_file = common.is_open_ci_env ? "27000@flexnet.trustedfirmware.org" : "7010@10.6.26.52:7010@10.6.26.53:7010@10.6.26.54:7010@10.6.26.56"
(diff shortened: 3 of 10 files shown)


arthur-she:
Test PR: https://github.com/arthur-she/mbedtls/pull/4
Test job:
  [PR head](https://ci.staging.trustedfirmware.org/view/Arthur%20MbedTLS/job/arthur-mbed-tls-pr-head/view/change-requests/job/PR-4-head/1/)
  [PR merge](https://ci.trustedfirmware.org/job/arthur-2-mbed-tls-pr-merge/view/change-requests/job/PR-4-merge/2/)

arthur-she:
Hi @yanesca ,
Could you please have a look?

Thank you

Arthur

vars/common.groovy
```diff
@@ -74,7 +77,7 @@ import groovy.transform.Field
 
 /* Maps platform names to the tag of the docker image used to test that platform.
  * Populated by init_docker_images() / gen_jobs.gen_dockerfile_builder_job(platform). */
-@Field static docker_tags = [:]
+@Field docker_tags = is_open_ci_env ? [('ubuntu-16.04'):'xenial', ('ubuntu-18.04'):'bionic'] : [:]
```
bensze01, line 80: Removing the `static` qualifier here will break the jobs on the internal CI - `gen_jobs.gen_dockerfile_builder_job()` will populate the wrong map if it is an instance field, rather than a class one.

A better solution would be to include this logic in `init_docker_images()` and populate the map there in the short term, while `gen_jobs.gen_dockerfile_builder_job()` doesn't work on OpenCI.

vars/mbedtls-release-Jenkinsfile
```diff
@@ -45,7 +45,8 @@ environ.set_tls_release_environment()
 
 timestamps {
     try {
-        common.init_docker_images()
+        if (!common.is_open_ci_env)
+            common.init_docker_images()
```
bensze01, line 49: If support for automatically updating docker images is not yet available on OpenCI, I think this logic would better fit inside `common.init_docker_images()`

bensze01 (changes requested):
Most of the changes seem reasonable, but removing the `static` qualifier will break our jobs on the internal CI.

vars/mbedtls-release-Jenkinsfile
arthur-she: Hi @bensze01 , @yanesca 
Thanks for your review. I will revise it.
Regarding the docker image, currently, for both TF-A and TF-M, we use a [Jenkins job](https://ci.trustedfirmware.org/job/ci-dockerfiles-deployment/) to monitor any merge activities on the [dockerfile repository](https://git.trustedfirmware.org/ci/dockerfiles.git/).
Gerrit will trigger the job to build the new docker image then push it to the [trustedfirmware](https://hub.docker.com/u/trustedfirmware) docker hub registry.
If we set up another Jenkins job, is it possible for you to create a webhook on this repository then the job can rebuild the docker image upon the latest dockerfile?

Thank you,

Arthur

gilles-peskine-arm: How do you test a Dockerfile update before making its image the official one?

bensze01: To expand a bit on Gilles's question:

Currently, we take advantage of the fact that multiple docker images can exist in parallel on our docker repo to test upcoming changes to the Dockerfiles.

If you create a branch on this repo, and point the parametrized job at your branch, `gen_jobs.gen_dockerfile_builder_job()` will recognize if the git hash of the Dockerfile changed, build the missing docker image and push it so that it can be tested.

This way the testing / update process for our Dockerfiles / images is the same as the one for our groovy scripts.

arthur-she: Hi @gilles-peskine-arm , @bensze01 ,
Thanks for the explanation.
I've revised the patch. Now we support automatically updating docker images on the Open CI.

Test job [here](https://ci.trustedfirmware.org/view/Mbed-TLS/job/mbed-tls-nightly-tests/57/consoleFull).

One question, how do you clean the stale images?

Thanks

arthur-she: To expand a bit on my question :)

What I am thinking about is that a new image will be created and pushed to the Docker registry once the dockerfile is changed.
After some time, the docker registry might contain many images which may no longer be used.

Will it happen?

bensze01: @arthur-she It's a fair point. I raised this with the internal CI team at the time, and they told me not to worry about it, since the cost to host the images was negligible.

I think we can leave it as is for now, then set up a cron job / some logic in existing jobs to clean out stale images once we have OpenCI up and running. In theory, (with proper locking) we could even simply delete any image older than x days, and any images that are actually used would get re-generated.

gilles-peskine-arm: > we could even simply delete any image older than x days, and any images that are actually used would get re-generated

I'd prefer not to do that. Regenerating an image can fail if a server it depends on is down or has moved. We've had this problem before. If old images automatically get deleted, we should have a way to mark some as persistent.

bensze01: Sure. It should be fairly simple to exclude the images that match the git hashes of the Dockerfiles on the HEAD of the master branch.

vars/gen_jobs.groovy
```diff
@@ -229,7 +231,7 @@ scripts/min_requirements.py --user
 #!/bin/sh
 set -eux
 ulimit -f 20971520
-export ARMLMD_LICENSE_FILE="7010@10.6.26.52:7010@10.6.26.53:7010@10.6.26.54:7010@10.6.26.56"
+export ARMLMD_LICENSE_FILE="${armlmd_license_file}"
```
gilles-peskine-arm, line 234: This overrides the setting in the Docker image. We did it that way because the license server moved at a time when a bug prevented us from updating the Docker images. This bug has been resolved but we never got around to removing the workaround. Please remove this line now, rather than building on it.

vars/gen_jobs.groovy
```diff
@@ -619,6 +622,9 @@ def gen_dockerfile_builder_job(platform, overwrite=false) {
     def dockerfile = libraryResource "docker_files/$platform/Dockerfile"
 
     def tag = "$platform-${common.git_hash_object(dockerfile)}"
+    def check_docker_image = "aws ecr describe-images --repository-name $common.docker_repo_name --image-ids imageTag=$tag"
+    def check_docker_image_open_ci = "docker manifest inspect $common.docker_repo:$tag > /dev/null 2>&1"
```
gilles-peskine-arm, line 626: Why not define `check_docker_image` based on `common.is_open_ci_env` here?

gilles-peskine-arm (changes requested):
On code inspection, this looks correct (for the Arm CI), but slightly overcomplicated in a couple of places.

Test runs on Arm CI:
* https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/360/ — ok
* https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-ci-testing/view/change-requests/job/PR-758-head/ — in progress
* https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-ci-testing/view/change-requests/job/PR-758-merge/ — in progress

I haven't reviewed the suitability for Open CI.

vars/gen_jobs.groovy
gilles-peskine-arm: https://github.com/ARMmbed/mbedtls-test/pull/31 is removing the workaround.

resources/docker_files/ubuntu-16.04/Dockerfile
```diff
@@ -117,7 +117,8 @@ RUN wget -q https://developer.arm.com/-/media/Files/downloads/compiler/DS500-PA-
 
 ENV ARMC5_BIN_DIR=/usr/local/ARM_Compiler_5.06u3/bin/
 ENV PATH=$PATH:/usr/local/ARM_Compiler_5.06u3/bin
-ENV ARMLMD_LICENSE_FILE=8225@armcc.licenses.mbedcloudtesting.com
+ARG ARMLMD_LICENSE_FILE=8225@armcc.licenses.mbedcloudtesting.com
```
bensze01, line 120: This license server has been dead for a while, we just never got around to removing it.
Please replace the default with the internal CI's values, as per [Gilles's comment](https://github.com/ARMmbed/mbedtls-test/pull/34/commits/54305920b3cb055a5766ba2a83ca10f1aafca114#r799656032).

```suggestion
ARG ARMLMD_LICENSE_FILE=7010@10.6.26.52:7010@10.6.26.53:7010@10.6.26.54:7010@10.6.26.56
```

vars/gen_jobs.groovy
```diff
@@ -627,18 +633,38 @@ def gen_dockerfile_builder_job(platform, overwrite=false) {
...
-                        sh """\
+                        if (common.is_open_ci_env) {
+                            withCredentials([string(credentialsId: 'DOCKER_AUTH', variable: 'TOKEN')]) {
+                                sh """\
+mkdir -p ${env.HOME}/.docker
+cat > ${env.HOME}/.docker/config.json << EOF
+{
+        "auths": {
+                "https://index.docker.io/v1/": {
+                        "auth": '${TOKEN}'
```
bensze01, line 651: Interpolating sensitive variables produces the following warning:
```
Warning: A secret was passed to "sh" using Groovy String interpolation, which is insecure.
		 Affected argument(s) used the following variable(s): [TOKEN]
		 See https://jenkins.io/redirect/groovy-string-interpolation for details.
```

So escape the dollar sign, and let the shell use the environment variable:
```suggestion
                        "auth": '\${TOKEN}'
```

bensze01 (changes requested):
Agree with Gilles, this looks good overall. Just one additional warning I'd like to fix on the OpenCI side.

resources/docker_files/ubuntu-16.04/Dockerfile
arthur-she: This change also applies to ubuntu-18.04/Dockfile, right?

vars/mbedtls-release-Jenkinsfile
arthur-she: Agree with you. Let's leave it as it is for now then find some way to clean up later.

resources/docker_files/ubuntu-16.04/Dockerfile
gilles-peskine-arm: Yes. Please note that [we're rewriting the 18.04 dockerfile to apply a number of changes we previously made to the 16.04 dockerfile](https://github.com/ARMmbed/mbedtls-test/pull/31). I think it would be easier to rebase your changes on top of #31 since #31 is an almost complete rewrite of the 18.04 dockerfile.

vars/gen_jobs.groovy
arthur-she: Done :)

vars/gen_jobs.groovy
arthur-she: Done :)

vars/gen_jobs.groovy
arthur-she: Done :)

resources/docker_files/ubuntu-16.04/Dockerfile
arthur-she: I can rebase my changes on top of #31, however, I saw there are several items on the TODO list.
Would it be better to do it after #31 get merged?

Thanks

gilles-peskine-arm: @arthur-she If you mean the investigation of test failures on 20.04, those are follow-ups: the objective of #31 is to have a docker image with 20.04, but we won't yet be running the tests on it. I think #31 just needs a few minor changes. I don't know how soon @minosgalanakis can make them though — Minos, when do you think you'll have time to rework #31?

gilles-peskine-arm: `outcomes.csv` is corrupted in a few places. It looks like a race condition where two concurrent threads were writing to the same file (either `outcomes.csv` itself, or one of the per-job files it's assembled from) and their writes got intermixed.

I've never seen this happen before. So either it's a very rare race condition that you were unlucky to be the first to discover (maybe due to different scheduling or load patterns on your Jenkins instance), or it's a race condition caused by your system having more concurrency somewhere.

I'll look a bit more closely to see if I can guess where the corruption happened.

vars/common.groovy
```diff
@@ -238,7 +238,8 @@ def maybe_notify_github(context, state, description) {
         description = description.take(MAX_DESCRIPTION_LENGTH - 1) + '…'
     }
 
-    githubNotify context: "${env.BRANCH_NAME} ${context}",
+    content = is_open_ci_env ? "TF OpenCI: ${env.BRANCH_NAME} ${context}" : "${env.JENKINS_URL} ${env.BRANCH_NAME} ${context}"
```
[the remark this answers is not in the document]
arthur-she: That's a good idea.
Done :)

vars/gen_jobs.groovy
```diff
@@ -627,18 +631,38 @@ def gen_dockerfile_builder_job(platform, overwrite=false) {
             node('dockerfile-builder') {
                 def image_exists = false
                 if (!overwrite) {
-                    def test_image_exists_sh = "aws ecr describe-images --repository-name $common.docker_repo_name --image-ids imageTag=$tag"
+                    def test_image_exists_sh = check_docker_image
                     image_exists = sh(script: test_image_exists_sh, returnStatus: true) == 0
```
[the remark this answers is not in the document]
arthur-she: Thanks for the suggestion.
Done

resources/docker_files/ubuntu-16.04/Dockerfile
gilles-peskine-arm: Scratch that, the explanation is a lot simpler. Some jobs ran out of disk space at various times. For example [`all_u16-test_no_pem_no_fs
`](https://ci.staging.trustedfirmware.org/job/arthur-mbed-tls-nightly-tests/14/execution/node/977/log/) ran out during the build, [`all_u16-test_ref_configs`](https://ci.staging.trustedfirmware.org/job/arthur-mbed-tls-nightly-tests/14/execution/node/916/log/) ran out when it tried to update a state file (and then failed to update the Jenkins log as well?), and some ran out while writing to their outcome file, resulting in missing parts including line breaks.

vars/common.groovy
```diff
@@ -230,7 +238,8 @@ def maybe_notify_github(context, state, description) {
         description = description.take(MAX_DESCRIPTION_LENGTH - 1) + '…'
     }
 
-    githubNotify context: "${env.BRANCH_NAME} ${context}",
+    content = is_open_ci_env ? "TF OpenCI: ${env.BRANCH_NAME} ${context}" : "Internal CI: ${env.JENKINS_URL} ${env.BRANCH_NAME} ${context}"
```
gilles-peskine-arm, line 241: `Internal CI: ${env.JENKINS_URL}` is redundant. But why not simply print `${env.JENKINS_URL}` in all cases?

bensze01: That's what it was before the most recent version - I'm leaning more towards removing the URL instead, to reduce visual noise / make it more helpful to contributors who might not now what the URLs mean.

It's a bit more obvious that they can see the "Open" CI logs, but not the internal ones.

resources/docker_files/ubuntu-16.04/Dockerfile
arthur-she: Yeah, I rebuilt the tests and got the "no space left on device" error. Trying to increase the volume size of our AMI now.

arthur-she: After increasing the volume size, I got the [build](https://ci.trustedfirmware.org/view/Mbed-TLS/job/mbed-tls-nightly-tests/85/console) successfully. 
I also rebased my changes.
Please have a look.

Thanks

vars/common.groovy
arthur-she: @gilles-peskine-arm , @bensze01 ,
Is there anything else I can help with for this PR?

arthur-she: Sorry, my fault.
I've removed the URL.

gilles-peskine-arm:
We've merged #31. Sorry it took us a bit longer than anticipated. There are conflicts in the Docker files now, so please rebase this pull request.

arthur-she:
Hi @gilles-peskine-arm ,
Thank you so much.
I've rebased my change.
Please have a look.

gilles-peskine-arm:
I'm afraid the rebase doesn't look right. You're re-applying some of Minos's changes that were later overridden. Can you please rebase your own changes on top of e1a61250a3ec7f6191044f9f7c510370efdd479f? The [commit list](https://github.com/ARMmbed/mbedtls-test/pull/34/commits) shouldn't have Minos's commits anymore.

arthur-she:
Sorry about that.
Since this branch is a mess and hard to rebase, I would prefer to close this pr and create another one.

PR #38, opened 2022-02-24
Condition Open CI changes based on Jenkins URL

arthur-she:
Modify the build script to fit two CIs, Trustedfirmware OpenCI and ARM internal CI.

### DIFF
--- a/vars/gen_jobs.groovy
+++ b/vars/gen_jobs.groovy
@@ -25,6 +25,8 @@ import groovy.transform.Field
 //Record coverage details for reporting
 @Field coverage_details = ['coverage': 'Code coverage job did not run']
 
+@Field armlmd_license_file = common.is_open_ci_env ? "27000@flexnet.trustedfirmware.org" : "7010@10.6.26.52:7010@10.6.26.53:7010@10.6.26.54:7010@10.6.26.56"
+
 def gen_simple_windows_jobs(label, script) {
     def jobs = [:]
 
@@ -333,6 +335,7 @@ def gen_windows_jobs(label_prefix='') {
 def gen_abi_api_checking_job(platform) {
     def jobs = [:]
     def job_name = "ABI-API-checking"
+    def credentials_id = common.is_open_ci_env ? "mbedtls-github-ssh" : "742b7080-e1cc-41c6-bf55-efb72013bc28"
 
     jobs[job_name] = {
         node('container-host') {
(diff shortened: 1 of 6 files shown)


gilles-peskine-arm:
Test jobs on Arm's Jenkins instance on 2f413d9f17177502aca1de360d5eb1dcd141abdb:
* [ ] [release job, development branch](https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/375/)
* [ ] [release job, mbedtls-2.28 branch](https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/376/)
* [ ] [release job, should have failures, with email notification](https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/377/) (https://github.com/ARMmbed/mbedtls-restricted/pull/759)
* [ ] [pr-head job, should pass](https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-ci-testing/view/change-requests/job/PR-758-head/4/) (https://github.com/ARMmbed/mbedtls-restricted/pull/758)
* [ ] [pr-merge job, should fail abi-break](https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-ci-testing/view/change-requests/job/PR-758-merge/6/) (https://github.com/ARMmbed/mbedtls-restricted/pull/758)

----

All failing because we've hit a rate limit on Docker. Need to try again later.

bensze01 (approved):
Looks good to me.

vars/gen_jobs.groovy
```diff
@@ -25,6 +25,8 @@ import groovy.transform.Field
 //Record coverage details for reporting
 @Field coverage_details = ['coverage': 'Code coverage job did not run']
 
+@Field armlmd_license_file = common.is_open_ci_env ? "27000@flexnet.trustedfirmware.org" : "7010@10.6.26.52:7010@10.6.26.53:7010@10.6.26.54:7010@10.6.26.56"
```
gilles-peskine-arm, line 28: This variable is no longer used. This was a temporary hack at a time we couldn't upgrade our Docker image. Now we're back to using the value set in the Dockerfile.

I don't mind whether the license file is decided inside or outside the Docker image, but let's pick one. Note that if it isn't set in the Docker image, we need to document how to set it when using the Docker image on a developer machine via `run.sh`.

gilles-peskine-arm (approved):
Looks good to me apart from the `ARMLMD_LICENSE_FILE` issue.

I'm marking this as approved because we can merge this now — only the armcc job will fail on OpenCI until this is fixed in a follow-up. Let me know what you prefer.

vars/gen_jobs.groovy
bensze01: @gilles-peskine-arm 
Actually this variable is now used only in the OpenCI-specific codepath on line 678:
```sh
docker build --build-arg ARMLMD_LICENSE_FILE=$armlmd_license_file -t $common.docker_repo:$tag .
```
I think the best solution would be to remove this line and copy `27000@flexnet.trustedfirmware.org` into that line directly:

```sh
docker build --build-arg ARMLMD_LICENSE_FILE=27000@flexnet.trustedfirmware.org -t $common.docker_repo:$tag .
```

That way we retain sensible defaults for developer machines.

Regardless, the code will work as-is, including the armcc jobs on both CIs.

PR #40, opened 2022-02-28
Introduce caching for base images and intermediate steps of docker image builds

bensze01:
This PR enables the caching of intermediate build steps that are identical between the current build and the most recent successful build for the same platform on both the Internal and the Open CI.

~~It also introduces caching for the base images we use on the internal CI. This works around the errors we are currently experiencing due to the pull limits imposed by Docker Hub.~~

I have filed an issue on Jira to increase our pull limits, but caching the base images on ECR may still be helpful once this is fixed.

**EDIT:** Our jenkins instance is now hooked up to the Arm-wide dockerhub cache, making the base image caching code unnecessary. I've removed the corresponding commit.

Test job run on [a branch](https://github.com/ARMmbed/mbedtls-test/tree/dev/bensze01/docker-caching-test) instrumented for testing (displays remaining pull limit):
https://jenkins-mbedtls.oss.arm.com/job/dockerfile-builder/178

### DIFF
--- a/vars/gen_jobs.groovy
+++ b/vars/gen_jobs.groovy
@@ -25,8 +25,6 @@ import groovy.transform.Field
 //Record coverage details for reporting
 @Field coverage_details = ['coverage': 'Code coverage job did not run']
 
-@Field armlmd_license_file = common.is_open_ci_env ? "27000@flexnet.trustedfirmware.org" : "7010@10.6.26.52:7010@10.6.26.53:7010@10.6.26.54:7010@10.6.26.56"
-
 def gen_simple_windows_jobs(label, script) {
     def jobs = [:]
 
@@ -661,7 +659,11 @@ def gen_dockerfile_builder_job(platform, overwrite=false) {
                     dir('docker') {
                         deleteDir()
                         writeFile file: 'Dockerfile', text: dockerfile
+                        def extra_build_args = ''
+
                         if (common.is_open_ci_env) {
+                            extra_build_args = '--build-arg ARMLMD_LICENSE_FILE=27000@flexnet.trustedfirmware.org'
+
                             withCredentials([string(credentialsId: 'DOCKER_AUTH', variable: 'TOKEN')]) {
                                 sh """\
 mkdir -p ${env.HOME}/.docker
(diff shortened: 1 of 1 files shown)


vars/gen_jobs.groovy
```diff
@@ -661,7 +659,11 @@ def gen_dockerfile_builder_job(platform, overwrite=false) {
                     dir('docker') {
                         deleteDir()
                         writeFile file: 'Dockerfile', text: dockerfile
+                        def extra_build_args = ''
+
                         if (common.is_open_ci_env) {
+                            extra_build_args = '--build-arg ARMLMD_LICENSE_FILE=27000@flexnet.trustedfirmware.org'
```
minosgalanakis, line 665: I am not certain why that change is needed. Ideally we want the licence server to not be included in the code base and provided as a jenkins secret using the `withCredentials()` method as displayed bellow. 

But it feels out of scope for this PR

vars/gen_jobs.groovy
```diff
@@ -675,17 +677,28 @@ cat > ${env.HOME}/.docker/config.json << EOF
...
                         }
+
+                        sh """\
+# Use BuildKit and a remote build cache to pull only the reuseable layers
+# from the last successful build for this platform
+DOCKER_BUILDKIT=1 docker build \
+    --build-arg BUILDKIT_INLINE_CACHE=1 \
+    $extra_build_args \
+    --cache-from $common.docker_repo:$platform-cache \
+    -t $common.docker_repo:$tag \
```
minosgalanakis, line 716: we could use the full notation of `--tag` instead of `-t` for consistency purposes

vars/gen_jobs.groovy
```diff
@@ -680,8 +680,29 @@ chmod 0600 ${env.HOME}/.docker/config.json
 """
                             }
                         } else {
+                            def base_image = platform.replaceFirst('-', ':')
                             sh """\
 aws ecr get-login-password | docker login --username AWS --password-stdin $common.docker_ecr
+
+# Pull cached base image
+# Ignore errors in case the cache doesn't exist yet
```
minosgalanakis, line 688: Wouldn't it make more sense to inverse the logic?

1. Attempt to pull from the cache
2. If cache fails then pull from the docker hub then update the cached base image

bensze01: That's basically what this code was doing, except it updated the base image even if the cache didn't fail.

Anyway, the jenkins instance has been hooked up to the Arm-wide dockerhub cache, so this commit is no longer necessary.

bensze01:
The Jenkins instance has been hooked up to the Arm-wide dockerhub cache, so the base image caching is no longer necessary.
I've removed the corresponding commit.

vars/gen_jobs.groovy
bensze01: The reason we settled on keeping it this way is because this lets (Arm) developers use the Dockerfiles as-is, without having to manually specify the license servers. The OpenCI codepath just overrides the Arm-internal defaults set in the Dockerfiles.

minosgalanakis (approved):
LGTM

gilles-peskine-arm (approved):
Looks good to me. In terms of fitness for purpose, I'm not familiar with Docker, BuildKit and the platforms we have available, so all I can say is that the code looks like what the documentation suggests and the CI logs look fine.

PR #47, opened 2022-05-17
Build required python sdists with default compiler

bensze01:
Calling `vcvarsall.bat` before `min_requirements.py` was interfering with installing python source distributions on Windows.

Example log for run without this patch, showing the bug:
https://ci.staging.trustedfirmware.org/view/Arthur%20MbedTLS/job/arthur-mbed-tls-nightly-tests/45/execution/node/109/log/

Example log of the equivalent step for a run with this patch applied:
https://ci.staging.trustedfirmware.org/view/Arthur%20MbedTLS/job/arthur-mbed-tls-nightly-tests/48/execution/node/1915/log/?consoleFull

### DIFF
--- a/vars/scripts.groovy
+++ b/vars/scripts.groovy
@@ -20,8 +20,8 @@
 import groovy.transform.Field
 
 @Field win32_mingw_test_bat = """\
-set CC=gcc
 if exist scripts\\min_requirements.py python scripts\\min_requirements.py || exit
+set CC=gcc
 if exist scripts\\make_generated_files.bat call scripts\\make_generated_files.bat || exit
 cmake . -G "MinGW Makefiles" || exit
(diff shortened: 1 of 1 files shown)


vars/scripts.groovy
```diff
@@ -20,8 +20,8 @@
 import groovy.transform.Field
 
 @Field win32_mingw_test_bat = """\
-set CC=gcc
 if exist scripts\\min_requirements.py python scripts\\min_requirements.py || exit
```
minosgalanakis, line 23: Given that this is common, would it be possible to add it outside of the '@Field' declarations? (Not a blocker, just curiosity)

minosgalanakis (approved):
Looks good,

daverodgman (approved):

PR #66, opened 2022-10-24
Driver-only hashes: test coverage in the CI

mprse:
Add CI test to verify if we have no regression in test coverage for driver-only-hashes build.

Resolves: https://github.com/Mbed-TLS/mbedtls/issues/6445
Related PR: https://github.com/Mbed-TLS/mbedtls/pull/6466 - this PR needs to be merged first.

### DIFF
--- a/vars/analysis.groovy
+++ b/vars/analysis.groovy
@@ -54,7 +54,8 @@ def gather_outcomes() {
                 }
                 try {
                     if (fileExists('tests/scripts/analyze_outcomes.py')) {
-                        sh 'tests/scripts/analyze_outcomes.py outcomes.csv'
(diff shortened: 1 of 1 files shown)


mpg (changes requested):
I disagree with the design. All the information about component to test and what to exclude should live in the main repo. We don't want to have to make a PR to mbedtls-test every time we change something. Actually, I don't think even an initial PR to mbedtls-test is necessary. Invoking `analyze_outcomes.py` with no arguments should run all the available tests. The ability to tune what it does on the command line should only be used for local testing.

mprse:
After changing the design this PR is not needed. Closing

PR #67, opened 2022-10-27
Report parallel errors to Github early

bensze01:
Update the status of the "TLS Testing" step when the first failure happens.

Test runs: [Internal CI][1] [Open CI][2]

In the Internal CI run, the patched script reported the initial failure 1 hour before the last parallel job finished, which is when the current script reports the errors.

In the Open CI run, the error was reported 1 hour 53 minutes earlier.

[1]: https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-ci-testing/view/change-requests/job/PR-759-merge/17/
[2]: https://ci.trustedfirmware.org/job/mbed-tls-restricted-pr-ci-testing/job/PR-759-merge/4/

### DIFF
--- a/vars/mbedtls.groovy
+++ b/vars/mbedtls.groovy
@@ -17,6 +17,25 @@
  *  This file is part of Mbed TLS (https://www.trustedfirmware.org/projects/mbed-tls/)
  */
 
+Map wrap_report_errors(Map jobs) {
+    return jobs.collectEntries { name, job ->
+        [(name): {
+            try {
+                job()
+            } catch (err) {
+                echo "Caught: ${err}"
+                echo "Failed job: ${name}"
+                if (!currentBuild.resultIsWorseOrEqualTo('FAILURE')) {
+                    currentBuild.result = 'FAILURE'
+                    common.maybe_notify_github 'TLS Testing', 'FAILURE',
+                            "Failures: ${name}…"
+                }
+                throw err
+            }
+        }]
+    }
+}
+
 def run_tls_tests(label_prefix='') {
     try {
         def jobs = [:]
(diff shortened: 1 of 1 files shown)


paul-elliott-arm (approved):
LGTM

gilles-peskine-arm:
Please make a test run that passes.

daverodgman (approved):

vars/mbedtls.groovy
```diff
@@ -17,6 +17,25 @@
...
+    return jobs.collectEntries { name, job ->
+        [(name): {
+            try {
+                job()
+            } catch (err) {
+                echo "Caught: ${err}"
+                echo "Failed job: ${name}"
+                if (!currentBuild.resultIsWorseOrEqualTo('FAILURE')) {
+                    currentBuild.result = 'FAILURE'
+                    common.maybe_notify_github 'TLS Testing', 'FAILURE',
```
gilles-peskine-arm, line 30: [There is a limit of 1000 statuses per sha and context](https://docs.github.com/en/rest/commits/statuses#create-a-commit-status) (sha=commit hash, context="TLS Testing"). This is one more update towards the limit, which is ok. I think currently there are 2: setting to pending, and final result, so we go from 500 runs to 333 runs before the status update saturates. That's still plenty except for the couple of PR we keep open for CI testing.

vars/mbedtls.groovy
```diff
@@ -17,6 +17,25 @@
...
+                            "Failures: ${name}…"
```
gilles-peskine-arm, line 31: This may be hard to distinguish from the final result which could look like this from GitHub's own truncation. We can tell from the `continuous-integration/jenkins/pr-merge` status, but down the line we might want to remove that because it's mostly noise. I don't think there's any reason today to have separate sub-statues:

* "Pre Test Checks" was for the transition period where some non-ancient PR didn't have a suitable `all.sh`, or for when the CI enforced some forbidden content, neither of which apply anymore.
* "TLS Testing" was separate from "Crypto Testing" back when there were two repositories, each with their own `all.sh`.
* "Result analysis" was separate because I didn't know how to report it properly, and [that's fixed](https://github.com/Mbed-TLS/mbedtls-test/pull/68).

gilles-peskine-arm (approved):
A couple of things to watch out, but we can work with what we have and improve later if it turns out to be a problem.

PR #70, opened 2022-11-02
Reduce the nightly merge activity window to 2 days

bensze01:


gilles-peskine-arm (approved):
2 days seems very short to me, but let's see how it goes.

PR #80, opened 2022-12-13
Increase timeout for all.sh jobs to 120 minutes

bensze01:
This matches the previous timeout for Windows jobs.

Signed-off-by: Bence Szépkúti <bence.szepkuti@arm.com>


paul-elliott-arm (approved):
LGTM - this timeout could potentially be lowered again by splitting up certain long running test suites as well.

PR #100, opened 2023-03-20
Fix the reported build result in timestamp files

bensze01:
The build result is only set to 'FAILED' if an exception propagates out of our groovy scripts, into Jenkins.

Set the result early, so that `archive_timestamps()` can report it correctly.


daverodgman (commented):
I don't see the difference in the outcomes? It reports SUCCESS IN both cases?

bensze01:
It should be fixed now. Added a new CI run above.

daverodgman (approved):

PR #102, opened 2023-04-06
Pr metrics dr

daverodgman:
Adding some fixes/updates to the pr-metrics scripts


pr-metrics/get-pr-data.py
```diff
@@ -3,8 +3,9 @@
 
 """Get PR data from github and pickle it."""
```
tgonzalezorlandoarm, line 4: Maybe we could change the comment as we are not using pickle anymore?

pr-metrics/get-pr-data.py
```diff
@@ -38,4 +39,13 @@
         p.update()
 
 with open("pr-data.p", "wb") as f:
-    pickle.dump(prs, f)
```
tgonzalezorlandoarm, line 41: I put this comment here as it does not let me put it in unchanged lines of code:
Nit: Maybe we can change `r = g.get_repo("ARMMbed/mbedtls")` from line 18 to `r = g.get_repo("Mbed-TLS/mbedtls")` ?

pr-metrics/get-pr-data.py
```diff
@@ -38,4 +39,13 @@
...
+    for i, p in enumerate(prs):
+        for retry in range(0, 9):
+            try:
+                g.dump(p, f)
+                break
+            except requests.exceptions.ReadTimeout:
+                delay = 2 ** retry
+                print(f"timeout; sleeping {delay} s and retrying...")
+                time.sleep(2 ** delay)
+        print(f"saved {i+1}/{len(prs)}")
```
tgonzalezorlandoarm, line 51: This could be printed even if it failed to save after the last retry. Also, in the case that the last retry fails, we are adding additional delay, which is the biggest of them all.

pr-metrics/prs.py
```diff
@@ -86,7 +91,7 @@ def is_community(pr):
 def quarter(date):
     """Return a string decribing this date's quarter, for example 19q3."""
```
tgonzalezorlandoarm, line 92: I think this comment should be changed now?

resources/windows/windows_testing.py
```diff
@@ -581,9 +582,8 @@ def test_visual_studio_built_code(self, test_run, solution_type):
                     git_worktree_path, test_run, vs_logger
                 )
             else:
-                solution_dir = os.path.join(
-                    git_worktree_path, "visualc", "VS2010"
-                )
+                solution_dir = glob.glob(os.path.join(
+                        git_worktree_path, "visualc", "VS*"))[0]
```
tgonzalezorlandoarm, line 586: Just in case: we are not expecting it to have more than one VS* folder, right? Otherwise this would bring a problem I think

tgonzalezorlandoarm (changes requested):

pr-metrics/get-pr-data.py
daverodgman: Reverted to using pickle, as this seems to work for me now.

PR #115, opened 2023-07-14
Support option to build the ecp with the new bignum interface

gabor-mezei-arm:
Support option to build the ecp with the new bignum interface.


gilles-peskine-arm:
Are we still doing this?

PR #119, opened 2023-07-26
Fix error reporting in case a branch is deleted while being tested

bensze01:
We report test failures to Github early, which causes Github to delete merge queue branches while they are still being tested. These deletions are reported to Jenkins via webhooks, which causes the `scm` variable to degenerate into a [NullSCM][1] object.

Handle such cases by extracting the needed information from `scm` early, and checking for the `NullSCM` type when performing checkouts.

This patch marks stages where we attempted to check out a NullSCM object as "unstable", which shows up as yellow on Blue Ocean. This helps quickly identify the  test that originally caused the failure (since its stage will show up as red).

~~At the time of writing, the stages on OpenCI turn yellow, as intended -  but the Internal CI needs a (fairly trivial) plugin upgrade for the stages to show up as yellow (UNSTABLE), rather than red (FAILED).~~

~~JIRA ticket for the Internal CI plugin upgrade: [OSSDEVOPS-6751][2]~~

[1]: https://javadoc.jenkins.io/hudson/scm/NullSCM.html
[2]: https://jira.arm.com/browse/OSSDEVOPS-6751

Test runs:
- [Internal CI merge queue Failure][3]
- [Open CI merge queue Failure][4]

[3]: https://jenkins-mbedtls.oss.arm.com/blue/organizations/jenkins/mbed-tls-pr-ci-testing/detail/gh-readonly-queue%2Fdev%2Fbensze01%2Fmerge-queue-test%2Fpr-7959-07c2bea18c0277686dc42d9d88ff0b6798700f29/1/pipeline/
[4]: https://mbedtls.trustedfirmware.org/blue/organizations/jenkins/mbed-tls-restricted-pr-ci-testing/detail/gh-readonly-queue%2Fdev%2Fbensze01%2Fmerge-queue-test%2Fpr-7959-07c2bea18c0277686dc42d9d88ff0b6798700f29/1/pipeline/

### DIFF
--- a/vars/checkout_repo.groovy
+++ b/vars/checkout_repo.groovy
@@ -17,9 +17,24 @@
  *  This file is part of Mbed TLS (https://www.trustedfirmware.org/projects/mbed-tls/)
  */
 
+import hudson.model.Result
+import hudson.scm.NullSCM
+import jenkins.model.CauseOfInterruption
+import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException
+
+void checkout_scm_not_null() {
+    if (scm instanceof NullSCM) {
+        echo 'scm is NullSCM - branch was deleted while being tested'
+        /* Color the stage yellow */
+        throw new FlowInterruptedException(Result.UNSTABLE, new CauseOfInterruption[0])
+    } else {
+        checkout scm
+    }
+}
+
 def checkout_repo() {
     if (env.TARGET_REPO == 'tls' && env.CHECKOUT_METHOD == 'scm') {
-        checkout scm
+        checkout_scm_not_null()
     } else {
         checkout_parametrized_repo(MBED_TLS_REPO, MBED_TLS_BRANCH)
     }
@@ -27,7 +42,7 @@ def checkout_repo() {
 
 def checkout_mbed_os_example_repo(repo, branch) {
     if (env.TARGET_REPO == 'example' && env.CHECKOUT_METHOD == 'scm') {
-        checkout scm
+        checkout_scm_not_null()
     } else {
         checkout_parametrized_repo(repo, branch)
     }
--- a/vars/checkout_repo.groovy
+++ b/vars/checkout_repo.groovy
@@ -22,53 +22,60 @@ import hudson.scm.NullSCM
 import jenkins.model.CauseOfInterruption
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException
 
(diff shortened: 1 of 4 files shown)


vars/checkout_repo.groovy
```diff
@@ -17,17 +17,32 @@
...
+import jenkins.model.CauseOfInterruption
+import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException
+
+void checkout_scm_not_null() {
+    if (scm instanceof NullSCM) {
+        echo 'scm is NullSCM - branch was deleted while being tested'
+        /* Color the stage yellow */
+        throw new FlowInterruptedException(Result.UNSTABLE, new CauseOfInterruption[0])
+    } else {
+        checkout scm
```
tom-cosgrove-arm, line 31: I don't know groovy, but is it like Java where you can have try-catch around a block? Would that help in case the branch is deleted while we are checking it out, or in case other errors happen during checkout?

bensze01: Yeah, Groovy is very much like Java.

The problem with `NullSCM` objects is that calling the `checkout` step on them is a no-op. `checkout` will complete without throwing any errors, and the build will eventually fail when it tries to eg. execute `all.sh`, which will be missing from the workspace.

I'm not exactly sure how the branch deletion events delivered to Jenkins cause `scm` do degenerate to a `NullSCM` object - but if we are worried about this happening between the `instanceof` check and the actual checkout, we could check the return value of the `checkout` step - a git checkout will return a map populated with information about the checked out commit, while a `NullSCM` checkout will return an empty map.

eg:

```groovy
if (!checkout(scm)) {
    echo "scm is $scm.class - branch was deleted while being tested"
    /* Color the stage yellow */
    throw new FlowInterruptedException(Result.UNSTABLE, new CauseOfInterruption[0])
}
```

If a non-null-scm checkout step fails due to eg. network connectivity problems / race conditions with the branch being deleted, it will throw an exception, which will propagate up the call stack. I'm not sure if we can diagnose these errors sufficiently to weed out deleted branches from other potential failures, but I can look into it.

tom-cosgrove-arm: > The problem with NullSCM objects is that calling the checkout step on them is a no-op

Ah, that makes a lot of sense, then.

In that case, no, the change is fine

paul-elliott-arm (approved):
LGTM

tom-cosgrove-arm (approved):
LGTM

PR #130, opened 2023-10-10
Prevent deadlock in api-abi-checking job

bensze01:
Calling `get_branch_information()` inside an `instrumented_node_job()` block causes the nesting of two `node()` blocks. This makes the script wait for a new node to be allocated on an executor, while still occpying the original executor as well. This slows the CI down considerably, and causes a deadlock if all nodes are occupied by api-abi checking jobs waiting for a node to be allocated.

Test runs:
- [Internal CI][1]
- [Open CI][2]

[1]: https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-ci-testing/view/change-requests/job/PR-906-merge/131/
[2]: https://mbedtls.trustedfirmware.org/job/mbed-tls-restricted-pr-ci-testing/view/change-requests/job/PR-906-merge/10/


gilles-peskine-arm (approved):
Looks plausible to me. Ok to merge if the CI passes.

tom-cosgrove-arm (approved):

PR #138, opened 2023-11-04
Flip the merge queue to OpenCI

gilles-peskine-arm:
We've switched again to making OpenCI the required CI, which runs on the merge queue. So we need to make OpenCI fake the status checks.

This is a minimal emergency fix. I've filed https://github.com/Mbed-TLS/mbedtls-test/issues/139 for more robustness.


gilles-peskine-arm:
Given that our normal merge process is blocked, I'm going ahead and merging this emergency one-character patch. If it doesn't work, we can revert it.

PR #144, opened 2024-01-02
speed up analyze-outcomes stage

daverodgman:
Improve perf of the analyse-outcomes stage of CI, which especially matters because it's a serial stage of the CI pipeline.

Currently, it takes around 247s to run xz on the CI.

Testing locally, current xz command takes 130s. With -0, it takes 15s (so expect ~28.5s on CI). This increases compressed size from 38 MB to 54 MB.

Using -T8 (8 threads) reduces it further to 2.6s locally (expect ~5s on CI). There's no penalty for going too big on thread count, but no benefit beyond about 8-16.

lzma, gzip, bzip2, zip are all worse.

Also, use grep instead of awk to extract failures, which saves a few more seconds.

CI run: https://jenkins-mbedtls.oss.arm.com/blue/organizations/jenkins/mbed-tls-pr-head/detail/PR-8530-head/9/pipeline/

### DIFF
--- a/vars/analysis.groovy
+++ b/vars/analysis.groovy
@@ -237,16 +237,16 @@ def process_outcomes() {
         deleteDir()
     }
 
-    // The complete outcome file is ~14MB compressed as I write.
+    // The complete outcome file is 2.1GB uncompressed / 56MB compressed as I write.
     // Often we just want the failures, so make an artifact with just those.
     // Only produce a failure file if there was a failing job (otherwise
     // we'd just waste time creating an empty file).
     if (gen_jobs.failed_builds) {
         sh '''\
-awk -F';' '$5 == "FAIL"' outcomes.csv >"failures.csv"
+grep ';FAIL;' outcomes.csv >"failures.csv"
 # Compress the failure list if it is large (for some value of large)
 if [ "$(wc -c <failures.csv)" -gt 99999 ]; then
-    xz failures.csv
+    xz -0 -T8 failures.csv
 fi
 '''
     }
@@ -258,7 +258,7 @@ fi
             }
         }
     } finally {
-        sh 'xz outcomes.csv'
+        sh 'xz -0 -T8 outcomes.csv'
         archiveArtifacts(artifacts: 'outcomes.csv.xz, failures.csv*',
                          fingerprint: true,
                          allowEmptyArchive: true)
--- a/vars/analysis.groovy
+++ b/vars/analysis.groovy
@@ -241,6 +241,9 @@ def process_outcomes() {
     // Often we just want the failures, so make an artifact with just those.
     // Only produce a failure file if there was a failing job (otherwise
     // we'd just waste time creating an empty file).
+    //
+    // Note that grep ';FAIL;' could pick up false positives, if another field such
+    // as test description or test suite was "FAIL".
     if (gen_jobs.failed_builds) {
         sh '''\
 grep ';FAIL;' outcomes.csv >"failures.csv"
--- a/vars/analysis.groovy
+++ b/vars/analysis.groovy
@@ -246,10 +246,10 @@ def process_outcomes() {
     // as test description or test suite was "FAIL".
     if (gen_jobs.failed_builds) {
         sh '''\
-grep ';FAIL;' outcomes.csv >"failures.csv"
+LC_ALL=C grep ';FAIL;' outcomes.csv >"failures.csv"
 # Compress the failure list if it is large (for some value of large)
 if [ "$(wc -c <failures.csv)" -gt 99999 ]; then
-    LC_ALL=C xz -0 -T8 failures.csv
+    xz -0 -T8 failures.csv
 fi
 '''
     }
--- a/vars/analysis.groovy
+++ b/vars/analysis.groovy
@@ -249,7 +249,7 @@ def process_outcomes() {
 LC_ALL=C grep ';FAIL;' outcomes.csv >"failures.csv"
 # Compress the failure list if it is large (for some value of large)
 if [ "$(wc -c <failures.csv)" -gt 99999 ]; then
-    xz -0 -T8 failures.csv
+    xz -0 -T0 failures.csv
 fi
 '''
     }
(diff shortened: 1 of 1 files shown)


vars/analysis.groovy
```diff
@@ -237,16 +237,16 @@ def process_outcomes() {
...
 
-    // The complete outcome file is ~14MB compressed as I write.
+    // The complete outcome file is 2.1GB uncompressed / 56MB compressed as I write.
     // Often we just want the failures, so make an artifact with just those.
     // Only produce a failure file if there was a failing job (otherwise
     // we'd just waste time creating an empty file).
     if (gen_jobs.failed_builds) {
         sh '''\
-awk -F';' '$5 == "FAIL"' outcomes.csv >"failures.csv"
+grep ';FAIL;' outcomes.csv >"failures.csv"
```
gilles-peskine-arm, line 246: This allows false positives if a test suite or a test case description is “FAIL”. I can live with that, but it requires a comment.

gilles-peskine-arm, line 246: GNU grep (at least some versions around the vintage we have on our CI) can be slow due to its handling of multibyte characters even when the pattern is pure ASCII. This only comes up in multibyte locales (e.g. `LC_CTYPE=en_US.UTF-8`). I think we have all locale categories set to `C` on the CI, but just in case, it would be more robust to explicitly set `LC_CTYPE` (or for that matter `LC_ALL`) to `C` when running `grep` and concerned about speed.

vars/analysis.groovy
```diff
@@ -237,16 +237,16 @@ def process_outcomes() {
...
 # Compress the failure list if it is large (for some value of large)
 if [ "$(wc -c <failures.csv)" -gt 99999 ]; then
-    xz failures.csv
+    xz -0 -T8 failures.csv
```
gilles-peskine-arm, line 249: Why `-T8`? AFAIK we don't have 8-core executors on the CI.

Beware of insisting on too much parallelism: this can create a load spike that slows down other executors, resulting in worse performance overall. You can't measure the impact just by running one job.

> Using -T8 (8 threads) reduces it further to 2.6s locally (expect ~5s on CI). There's no penalty for going too big on thread count, but no benefit beyond about 8-16.

Measuring on a machine with a different number of cores is not useful to guess at the performance on the CI.

gilles-peskine-arm, line 249: > lzma, gzip, bzip2, zip are all worse.

I'd expect them to result in worse compression, but maybe we're ok with less compression but less CPU consumption?

gilles-peskine-arm (changes requested):
OK to switch to grep, but only if the limitation is documented.

Not ok to use more threads than we have cores.

vars/analysis.groovy
daverodgman: The trade-off looked non-favourable in all cases; the closest was probably gzip, which is faster (in single-thread mode, but it doesn't have a multi-thread mode), but almost twice as big output - so `xz -T<n>` wins overall.

gilles-peskine-arm: Multithreaded mode may be counterproductive anyway, but have you tried `pigz`? (In a realistic scenario, not on your machine!)

daverodgman: `pigz -1` does look considerably better for both perf (0.36s locally) and size (25 MB, about 2x better).

gilles-peskine-arm: By the way, we need to keep in mind that another element of the compromise is storage. That's the reason we compress, not to save on download time/bandwidth. A vast majority of outcome files are never consumed, but all are stored for some time. I don't know how much we're spending on storage, and in particular how the (compressed) outcome files compare with logs.

daverodgman: We can always up the storage budget if we need to. As it stands, an extra 18 MB / PR run is unlikely to cause big budget issues IMO.

daverodgman: I'm tracking the `pigz` suggestion as a separate follow-up, since that requires additional CI work beyond the scope of this PR.

tom-cosgrove-arm (dismissed):

vars/analysis.groovy
```diff
@@ -249,7 +249,7 @@ def process_outcomes() {
 grep ';FAIL;' outcomes.csv >"failures.csv"
 # Compress the failure list if it is large (for some value of large)
 if [ "$(wc -c <failures.csv)" -gt 99999 ]; then
-    xz -0 -T8 failures.csv
+    LC_ALL=C xz -0 -T8 failures.csv
```
gilles-peskine-arm, line 252: `LC_ALL=C` was supposed to apply to `grep`, not `xz`.

vars/analysis.groovy
gilles-peskine-arm: I still don't see a reason to use `-T8`. Why not `-T0`?

As far as I understand, the actual number of threads is limited by the memory consumption (which I don't expect to matter at `-O0`), by the `-T` value and by the file size, but not by the number of cores. Looking at the `xz` source code, `lzma_cputhreads()` is called to determine what `hardware_threads_get()` will return when using `-T0`, and then the encoder only relies on `hardware_threads_get()` and not `lzma_cputhreads()`.

Since 8 is an arbitrary number here, I'd want to see some benchmark that justifies it. Otherwise please use `-T0` (the number of cores seems like a sensible default choice).

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
CI is green

@gilles-peskine-arm are you happy enough with this now to approve?

vars/analysis.groovy
gilles-peskine-arm: This is still unaddressed.

daverodgman:
@gilles-peskine-arm I have updated to use `-T0`, please review

tom-cosgrove-arm (approved):

gilles-peskine-arm (approved):
Looks good to me on code inspection.

Ok to merge with a passing CI run (either release or PR will do).

daverodgman:
Test run shows xz taking 15s https://jenkins-mbedtls.oss.arm.com/blue/rest/organizations/jenkins/pipelines/mbed-tls-pr-head/branches/PR-8530-head/runs/12/nodes/8129/log/?start=0

PR #157, opened 2024-02-09
Use the MSVC versions supported by each branch

bensze01:
2.28 PRs test MSVC 2013, and nightlies test 2013, 2015 and 2017.

development PRs and nightlies only test MSVC 2017.

### DIFF
--- a/vars/common.groovy
+++ b/vars/common.groovy
@@ -307,11 +307,19 @@ def check_every_all_sh_component_will_be_run(BranchInfo info) {
 }
 
 def get_supported_windows_builds() {
-    def vs_builds = []
+    def vs_builds
     if (env.JOB_TYPE == 'PR') {
-        vs_builds = ['2013']
+        if (env.CHANGE_TARGET == 'mbedtls-2.28') {
+            vs_builds = ['2013']
+        } else {
+            vs_builds = ['2017']
+        }
     } else {
-        vs_builds = ['2013', '2017']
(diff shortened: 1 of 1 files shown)


vars/common.groovy
```diff
@@ -307,11 +307,19 @@ def check_every_all_sh_component_will_be_run(BranchInfo info) {
...
     if (env.JOB_TYPE == 'PR') {
-        vs_builds = ['2013']
+        if (env.CHANGE_TARGET == 'mbedtls-2.28') {
+            vs_builds = ['2013']
+        } else {
+            vs_builds = ['2017']
+        }
     } else {
-        vs_builds = ['2013', '2017']
+        if (env.MBED_TLS_BRANCH == 'mbedtls-2.28') {
```
gilles-peskine-arm, line 318: How is `MBED_TLS_BRANCH` determined?

I fear that this approach will break if we make bugfix branches. Does it even work with security fixes that target `mbedtls-2.28-restricted`?

What we've done in the past was to read a file in the repository and base the testing on that. For example, at one point we had C89 branches tested with VS2010 and C99 branches tested with VS2013. See d4272aca6193aa7a2d8ec03b8437a0dc00b0dc54 and 5b1eb91d4b2c703a44e1c49880ed927bf1b9f530.

By the way, that second commit suggests that we might need to change `windows_testing.py` too?

bensze01:
Superceded by #202

PR #161, opened 2024-03-28
Increase linux testing timeout

bensze01:
We've breached 2 hours in  `all_u16-release_test_valgrind_psa`, so increase the timeout value.

With the current implementation of "silent logging", timeouts also destroy all the logs, so turn verbose logs on for now.

Test run: https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/587/


tom-cosgrove-arm (approved):
LGTM

PR #165, opened 2024-04-25
Add coverity push script

paul-elliott-arm:
## Description

This is the script to allow us to push coverity builds from Jenkins (or anyone's machine come to that)

Functionality is documented within the script, but to test: 

either:

```
export COVERITY_TOKEN=token
path/to/mbedtls/scripts/push_coverity_scan.py path/to/mbedtls -b origin/coverity_scan -e email@arm.com
```
or 
```
path/to/mbedtls/scripts/push_coverity_scan.py path/to/mbedtls -t token -b origin/coverity_scan -e email@arm.com
```

The token can be found from the coverity web interface, and for the email address I would use your own email address for now (email address is unfortunately compulsary). Conversations are ongoing about which address to use for notifications in the future, the notification email will look like this:

```
 Your request for analysis of ARMmbed/mbedtls has been completed successfully.
    The results are available at <url>

    Build ID: 607763

    Analysis Summary:
       New defects found: 0
       Defects eliminated: 0
```

Should new defects be found, another email will be generated, but this will go to the whole team (this is specified within coverity itself)

WARNING - the mbedtls directory specified will be switched to the latest version of the branch specified, any outstanding differences will currently cause the script to fail (in my opinion better than forcing anything).

This constitutes stage 1 of getting coverity push back into CI, getting this script reviewed so that we can at least check its in the correct shape, then stage 2 will be adding the groovy and testing in the repo. I currently envisage this to be added shortly after the existing push from `development` to `coverity_scan` branches in the nightly runs.

### DIFF
--- a/coverity/push_coverity_scan.py
+++ b/coverity/push_coverity_scan.py
@@ -1,0 +1,454 @@
+#!/usr/bin/env python3
+
+""" Build mbedtls using the coverity toolset and upload said build to coverity.
+
+A small script designed to be run both in the CI and on local users machines.
+
+Required:
+
+1. Path to mbedtls directory
+2. A project coverity token (got from the Coverity site / Project Settings
+3. An email address to send notifications to.
+
+(The last of these is annoying, but things will not work without it, all that it
+is used for is to notify that the build was submitted)
+
+The token can either be passed in as an argument with -t, or via the environment
+in the variable 'COVERITY_TOKEN'
+
+Other options:
+
+* -b / --branch: If specified, this branch will be checked out prior to build.
+* -c / --covtools: If specified, the coverity tools will be downloaded here. If
+    there is already a set of coverity tools in the specified directory, they
+    will be checked for appropriate version, and overwritten only if necessary.
+* -p / --pre-build-step: Specify the command to run pre-build - defaults to
+    'make clean'.
+* -s / --build-step: Specify the command to run to build the project - defaults
+    to 'make -j'.
+
+* -l / --log: Specify a file to log information on the build to.
+* -m / --backupdir: If specified, this will be used as a directory to backup
+    built tar files to.
+* -v / --verbose: If specified, all logging will be done to stdout.
+
+"""
+
+# Copyright The Mbed TLS Contributors
+# SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
+
+import argparse
+from subprocess import Popen, PIPE
+import shlex
+from traceback import format_exc
+
+from typing import Tuple, Iterable
+
+import os
+import pathlib
+import shutil
+from tempfile import mkstemp, mkdtemp
+from datetime import datetime
+import tarfile
+import hashlib
+import logging
+import sys
+
+import requests
+
+class BuildError(Exception):
+    """ Exception class for build errors """
+    pass
+
+
+class ConfigError(Exception):
+    """ Exception class for configuration errors """
+    pass
+
+
+def do_shell_exec(exec_string: str, expected_result: int = 0) -> Tuple[bool, str, str]:
+
+    """ Execute the given string in a shell, try to ascertain success
+    """
+
+    shell_process = Popen(shlex.split(exec_string), stdin=PIPE, stdout=PIPE, stderr=PIPE)
+
+    (shell_stdout, shell_stderr) = shell_process.communicate()
+
+    if shell_process.returncode != expected_result:
+        return False, shell_stdout.decode("utf-8"), shell_stderr.decode("utf-8")
+
+    return True, shell_stdout.decode("utf-8"), shell_stderr.decode("utf-8")
+
+def check_coverity_scan_tools_version(token: str, tools_dir: str) -> bool:
+
+    """ Get the md5 of the coverity tools package from coverity, so we can check we have the latest
+        version.
+    """
+
+    post_data = [('token', token),
+                 ('project', 'ARMmbed/mbedtls'),
+                 ('md5', '1')]
+
+    # Presumption of linux here, could support other build types?
+    md5_request = requests.get('https://scan.coverity.com/download/linux64', data=post_data,
+                               timeout=60)
+    md5_request.raise_for_status()
+
+    md5_path = pathlib.Path(tools_dir)
+    md5_path = md5_path / 'coverity_tool.md5'
+
+    with md5_path.open('r') as md5_file:
+        tools_hash = md5_file.read()
+
+    return tools_hash == md5_request.text
+
+def backup_config_files(mbedtls_dir: pathlib.Path, restore: bool) -> None:
+
+    """Backup / Restore config file."""
+
+    config_path = pathlib.Path(mbedtls_dir)
+    config_path = config_path / 'include' / 'mbedtls' / 'mbedtls_config.h'
+    config_path.resolve()
+
+    backup_path = config_path.with_suffix('.h.bak')
+
+    if restore:
+        if backup_path.is_file():
+            backup_path.replace(config_path)
+    else:
+        shutil.copy(config_path, config_path.with_suffix('.h.bak'))
+
+def filter_root_tar_dir(tar_file: tarfile.TarFile) -> Iterable[tarfile.TarInfo]:
+
+    """ This function allows extraction of the contents of the first containing directory in the tar
+        directly to the target folder, by stripping the first containing folder from each path.
+    """
+
+    for tar_member in tar_file.getmembers():
+
+        member_path = pathlib.Path(tar_member.path)
+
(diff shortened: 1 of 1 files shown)


paul-elliott-arm:
One possible improvment (although not yet required) would be to parameterise the project so that we can easily use this on other projects, however I'm not sure how many coverity scan projects we are allowed to have!

gilles-peskine-arm (changes requested):
I've read the code, but not tried to run it. This looks good from a high-level view, but there are several places where I found the code hard to follow. Some places seem to reinvent standard library functions, and some places have difficult-to-understand error handling that looks like it could be simpler. Generally, in Python, prefer context handlers (`with …`) if possible.

coverity/push_coverity_scan.py
```diff
@@ -0,0 +1,454 @@
...
+
+    logger.log(logging.INFO, "Downloading Coverity Scan....")
+
+    # Presumption of linux here, could support other build types?
+    package_request = requests.get('https://scan.coverity.com/download/linux64', data=post_data,
+                                   timeout=60)
+    package_request.raise_for_status()
+
+    # Write this to a temp file, as we need to extract it
+    temp_file_handle, temp_file_name = mkstemp()
```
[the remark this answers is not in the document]
paul-elliott-arm: As it turns out, no we don't.

coverity/push_coverity_scan.py
```diff
@@ -0,0 +1,454 @@
...
+    """ Exception class for build errors """
+    pass
+
+
+class ConfigError(Exception):
+    """ Exception class for configuration errors """
+    pass
+
+
+def do_shell_exec(exec_string: str, expected_result: int = 0) -> Tuple[bool, str, str]:
```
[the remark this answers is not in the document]
paul-elliott-arm: Using subprocess.run() as this is now the standard function apparently.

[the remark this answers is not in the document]
gilles-peskine-arm: I strongly prefer using the `check_call` or `check_output` functions when possible, because they make it immediately obvious that a failure will raise an exception. `subprocess.run` is unfortunately has an unsafe default `check=False`, so it's preferrable to `subprocess.Popen` but not to the check functions.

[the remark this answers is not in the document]
paul-elliott-arm: Function removed.

coverity/push_coverity_scan.py
```diff
@@ -0,0 +1,454 @@
...
+
+    finally:
+        # Clean up, if necessary
+        if args.backupdir is None and tar_file_set:
+            os.unlink(tar_file)
+
+        if args.covtools is None and tools_path_set:
+            shutil.rmtree(tools_path)
+
+        backup_config_files(mbedtls_path, True)
```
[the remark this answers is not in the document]
paul-elliott-arm: Its the same disease most of our scripts have which backup config.h, so I would argue this isn't something we can do a whole lot about.

coverity/push_coverity_scan.py
```diff
@@ -0,0 +1,417 @@
...
+        logger.log(logging.INFO, result.stdout.decode("utf-8"))
+
+        result = run(['git', 'submodule', 'update', '--init'], capture_output=True, check=True)
+        logger.log(logging.INFO, result.stdout.decode("utf-8"))
+
+
+    # Backup config files here prior to running config.py, as the branch checkout may also have
+    # changed them, and backing up before this point will end up with old versions of the file being
+    # restored.
+    backup_config_files(logger, mbedtls_path, False)
```
gilles-peskine-arm, line 182: `mbedtls_path` is not defined here. There's a variable `mbedtls_dir` but we're already chdir'ed into that, so I guess `os.curdir`? Because of the `chdir`, using `mbedtls_dir` here would be wrong if it's a relative path.

coverity/push_coverity_scan.py
```diff
@@ -0,0 +1,417 @@
...
+2. A project coverity token (got from the Coverity site / Project Settings
+3. An email address to send notifications to.
+
+(The last of these is annoying, but things will not work without it, all that it
+is used for is to notify that the build was submitted)
+
+The token can either be passed in as an argument with -t, or via the environment
+in the variable 'COVERITY_TOKEN'
+
+Other options:
```
gilles-peskine-arm, line 19: It might be better to leave out the list of options here. They're already documented through the `argparse` calls and I fear that we'll forget to update the text here.

coverity/push_coverity_scan.py
```diff
@@ -188,58 +153,53 @@ def download_coverity_scan_tools(logger: logging.Logger, token: str, tools_dir:
...
-
-    """ Build mbedtls located in the passed in dir, using the tools specified, using the given
-        pre-build and build commands. Tar the results up into the given file name, as required by
-        coverity. """
+    Build the MBed TLS source located in the passed in dir, using the tools specified, using the
+    given pre-build and build commands. Tar the results up into the given file name, as required by
+    coverity.
+    """
 
     os.chdir(mbedtls_dir)
```
gilles-peskine-arm, line 163: Because of the `chdir` call here, I'm not sure how the each command line arguments will be interpreted if they're relative paths, or if they even _can_ be relative paths. It would be more robust to not call `chdir` and pass `cwd=mbedtls_dir` to `subprocess.run` calls.

gilles-peskine-arm (changes requested):
Thanks for the updates! On code inspection, I believe I now understand what the code is doing, and would be able to maintain it going forward.

I haven't tried running the script yet (and there's a typo that looks like it would fail midway).

I wonder how to test it: how far can we go before consuming a limited Coverity credit: is that consumed by the build request? By triggering? Could we have a test mode in the script that does everything except the credit-consuming step, and leaves behind or prints out all necessary information to do the credit-consuming step manually?

paul-elliott-arm:
> I wonder how to test it: how far can we go before consuming a limited Coverity credit: is that consumed by the build request? By triggering? Could we have a test mode in the script that does everything except the credit-consuming step, and leaves behind or prints out all necessary information to do the credit-consuming step manually?

I have added `-n/--no-upload` which will cover this case. Should you use a backup directory for the generated tar file (rather than a temp file) this tar file can then be manually uploaded later, or you can just use this to test the tools download / build steps.

coverity/push_coverity_scan.py
paul-elliott-arm: I'm really unsure on this one, the scripts themselves generally assume being run from the root MBedTLS dir, so this is how I made it, I haven't noticed any issues up till now.

gilles-peskine-arm (dismissed):
I ran the script locally with `--no-upload` and it worked, producing a plausible-looking `mbedtls-24-12-06.tar.gz` in the `--backupdir` directory. So this looks good to me both on code reading and on execution apart from possibly the upload step. I'd still like to check that the upload step works for me while you're still partly around, but not today, and in the meantime I can approve this PR already, we can take it from there.

coverity/push_coverity_scan.py
```diff
@@ -0,0 +1,441 @@
...
+            # If we want a backup of the tar file, then make one.
+            if args.backupdir is not None:
+                backup_path = pathlib.Path(args.backupdir)
+
+                if not backup_path.is_dir():
+                    logger.log(logging.INFO, 'Backup dir does not exist, creating.')
+                    backup_path.mkdir()
+
+                backup_path = backup_path.resolve()
+                backup_path = backup_path / datetime.today().strftime('mbedtls-%y-%m-%d.tar.gz')
```
gilles-peskine-arm, line 411: Non-blocker: a 2-digit year??
```suggestion
                backup_path = backup_path / datetime.today().strftime('mbedtls-%Y-%m-%d.tar.gz')
```

gilles-peskine-arm (changes requested):
LGTM apart from documentation that's slightly out of date.

I successfully submitted a build with
```
coverity/push_coverity_scan.py --backupdir ~/OUTGOING/coverity --branch dev --covtools ~/INCOMING/covtools --email … --token $(…) -v ~/z/dev/mbedtls/main
```

I'd like a few minor usability improvements, but I'd prefer to merge this first and then make a new pull request.

coverity/push_coverity_scan.py
```diff
@@ -0,0 +1,441 @@
+#!/usr/bin/env python3
+
+""" Build Mbed TLS using the coverity toolset and upload said build to coverity.
+
+A small script designed to be run both in the CI and on local users machines.
+
+Required:
+
+1. Path to Mbed TLS directory
+2. A project coverity token (got from the Coverity site / Project Settings
```
gilles-peskine-arm, line 10: Apparently you need to have “Maintainer/Owner” permission, not just “Contributor/Member”.

PR #175, opened 2024-09-17
Bypass Github rate limits on anonymous git clones

bensze01:
This PR bypasses Github's rate limit on anonymous git clones by performing all clones and fetches, including submodule clones using SSH (authenticated), and reduces the network traffic and time taken for checkouts by making all clones and fetches shallow.

Test runs: TODO


vars/checkout_repo.groovy
```diff
@@ -66,7 +66,23 @@ Map<String, String> checkout_repo() {
     } else {
         scm_config = parametrized_repo(env.MBED_TLS_REPO, env.MBED_TLS_BRANCH)
     }
-    return checkout_report_errors(scm_config)
+
+    // Use bilingual scripts when manipulating the git config
+    def sh_or_bat = isUnix() ? {args -> sh(args)} : {args -> bat(args)}
+    // Set global config so its picked up when cloning submodules
+    sh_or_bat 'git config --global "url.git@github.com:.insteadOf" "https://github.com/"'
```
minosgalanakis, line 73: Is this a workaround or a permanent fix? Why are we hardcoding something which could be provided by the parameters or the enviroment?  e.g `env.MBED_TLS_REPO`

vars/checkout_repo.groovy
```diff
@@ -90,13 +106,12 @@ Map<String, Object> parametrized_repo(String repo, String branch) {
...
         userRemoteConfigs: [[
             url: repo,
-            refspec: '+refs/heads/*:refs/remotes/origin/* +refs/pull/*:refs/pull/*',
             credentialsId: env.GIT_CREDENTIALS_ID
         ]],
         branches: [[name: branch]],
         extensions: [
-            [$class: 'CloneOption', timeout: 60],
-            [$class: 'SubmoduleOption', recursiveSubmodules: true],
+            [$class: 'CloneOption', timeout: 60, honorRefspec: true, noTags: true, shallow: true],
```
minosgalanakis, line 113: Is there any reason we are not checking out tags? They are not that many in  total and do not consume much bandwidth?

They will be required in the future by the release preparation scripts.

minosgalanakis (commented):
Looks good overall, just had two minor comments.

vars/checkout_repo.groovy
bensze01: It's a permanent fix. It's needed because the URL of the submodules (specified inside the Mbed-TLS repo's commits) is an https url, which trips Github's anonymous clone rate limits.

The only thing this config option does is redirect clones/fetches from `https://github.com/*` to `ssh://git@github.com/*`, which always point to the same repo - the `*` part is still totally under the control of `env.MBED_TLS_REPO`.

bensze01:
@minosgalanakis I think I've addressed your review feedback.

minosgalanakis (dismissed):
LGTM

vars/checkout_repo.groovy
eleuzi01, line 73: Correct me if I'm wrong but I think this should be:
```suggestion
    sh_or_bat 'git config --global url."git@github.com:".insteadOf "https://github.com/"'
```

bensze01: Both of those are equivalent.
During quote removal any characters "touching" the quote marks will be prepended/appended to the string inside the quote marks.

So the arugments passed to `git` will be  `argv = ['git', 'config', '--global', 'url.git@github.com:.insteadOf', 'https://github.com/']` either way.

tom-cosgrove-arm: Since there aren't any spaces or other shell special characters in there, why are the double quotes needed at all?

bensze01: True, they are superfluous. Will go ahead and remove them.

eleuzi01 (approved):
LGTM

minosgalanakis (approved):
LGTM

PR #182, opened 2024-11-06
Fix sending test-result E-mails

bensze01:
This call was missed / accidentally dropped during a rebase of commit 8d9b2b78df05a3bd0ef90e4cfe021373fc65c652.

Also initialize infos at declaration, so an early failure doesn't result in a null pointer dereference in the finally block.

Test runs:
- Internal CI
  - [E-mail set][1]
  - [E-mail unset][2]
- OpenCI
  - [E-mail set][3]

[1]: https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/727/
[2]: https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/728/
[3]: https://mbedtls.trustedfirmware.org/job/mbedtls-release-ci-testing/287/

### DIFF
--- a/vars/common.groovy
+++ b/vars/common.groovy
@@ -417,7 +417,9 @@ void send_email(String name, Collection<BranchInfo> infos) {
     String branches = infos*.branch.join(',')
     def failed_builds = infos.collectMany { info -> info.failed_builds}
     String coverage_details = infos.collect({info -> "$info.branch:\n$info.coverage_details"}).join('\n\n')
-    if (failed_builds) {
+
+    boolean failed = infos.size() == 0 || failed_builds.size() > 0
+    if (failed) {
         failures = failed_builds.join(", ")
         emailbody = """
 $coverage_details
@@ -436,7 +438,7 @@ Logs: ${env.BUILD_URL}
         recipients = env.TEST_PASS_EMAIL_ADDRESS
     }
     subject = ((is_open_ci_env ? "TF Open CI" : "Internal CI") + " ${name} " + \
-           (failed_builds ? "failed" : "passed") + "! (branches: ${branches})")
+           (failed ? "failed" : "passed") + "! (branches: ${branches})")
     echo """\
 Subject: $subject
 
--- a/vars/common.groovy
+++ b/vars/common.groovy
@@ -413,7 +413,7 @@ done
     )
 }
 
-void send_email(String name, Collection<BranchInfo> infos) {
+void maybe_send_email(String name, Collection<BranchInfo> infos) {
     String branches = infos*.branch.join(',')
     def failed_builds = infos.collectMany { info -> info.failed_builds}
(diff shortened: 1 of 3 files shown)


gilles-peskine-arm (dismissed):
Looks good to me on code reading.

paul-elliott-arm (dismissed):
LGTM (Certainly makes sense anyway)

bensze01:
Pushed a commit enabling testing the email mechanism, plus a small logic fix and a drive-by improvement of `send_email()` I made while working on the former.

paul-elliott-arm (commented):
LGTM at least by inspection (provided successful test runs)

vars/common.groovy
```diff
@@ -435,14 +438,19 @@ Logs: ${env.BUILD_URL}
 """
         recipients = env.TEST_PASS_EMAIL_ADDRESS
```
gilles-peskine-arm, line 439: So now the job configuration and/or runtime parameters determine whether to send an email. I like that. (In addition to being more robust, it means we can fire a job and get an email notification when it's finished, if we care.) Do we need to revise the configuration of some jobs to start offering the `TEST_xxx_EMAIL_ADDRESSES` environment variable in parametrized builds, or to make them empty by default? For example [`mbedtls-restricted-release-new` on OpenCI lists `TEST_FAIL_EMAIL_ADDRESS`](https://review.trustedfirmware.org/plugins/gitiles/ci/mbedtls/mbed-tls-job-configs/+/refs/heads/master/mbedtls-restricted-release-new.yaml), but AFAICT it's a hard-coded list, not a property we can change.

gilles-peskine-arm (changes requested):
We've lost a bit of useful logging, and I'm not sure if the job configurations need changing.

vars/common.groovy
bensze01: It is already a parameter on the internal CI. Will push a gerrit change that parametrizes it on the OpenCI as well.

gilles-peskine-arm (dismissed):
Looks good to me on code inspection, assuming we update the relevant job configurations to make `TEST_{PASS,FAIL}_EMAIL_ADDRESSES` environment variables with suitable defaults.

gilles-peskine-arm (approved):
LGTM (pending test jobs being ok, and assuming the necessary configuration changes are done)

vars/common.groovy
bensze01: Job config change to parametrize this on the OpenCI:
https://review.trustedfirmware.org/c/ci/mbedtls/mbed-tls-job-configs/+/33347

Not sure whether all of the default e-mail addresses are correct/active, so I've left them in place for now where they were effectively being used prior to this PR.

bensze01:
All CI runs passed, config change merged.

PR #184, opened 2024-11-13
Fix result analysis in failed multi-branch test runs

### LINKED ISSUE #127: Show analyze_outcomes step on the BlueOcean view
On Jenkins, on the BlueOcean view, the list of sub-steps of the `result-analysis` appears as a long list of “Restore files previously stashed”. Those are boring steps we don't care about. There's more than 100 of them, and BlueOcean truncates the list at 100. This causes the one interesting sub-step of `result-analysis` to be hidden, namely, the one that runs `analyze_outcomes.py`. Thus, when `analyze_outcomes.py` fails, this causes a failure of `result-analysis` that cannot be explained if you only look at BlueOcean and not the classic pipeline view.

Definition of done for this issue: in the BlueOcean view, the report of `result-analysis` makes the interesting steps visible (both on success and on failure). Whether the individual “Restore files previously stashed” steps are visible at all doesn't matter since those are very unlikely to matter.

bensze01:
This PR fixes several issues that occur when the name of the Mbed TLS branch being tested is included in file or stash names. This happens when we test multiple branches at once, eg. in framework PRs or if multiple branches are specified in a parameterized job.

In addition, the second commit breaks gen_docker_job into sub-stages, so that the result analysis display on Blue ocean is not overwhelmed by the unstash stages. 

Fixes #127 

Test runs:
 - development: good + skipped/failed tests
   - [Internal CI][1]
   - [OpenCI][2]
 - development: release job
   - [OpenCI](https://mbedtls.trustedfirmware.org/job/mbedtls-release-ci-testing/297/)
 
[1]: https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-test-parametrized/372/
[2]: https://mbedtls.trustedfirmware.org/job/mbed-tls-restricted-pr-test-parametrized/64/

### DIFF
--- a/vars/analysis.groovy
+++ b/vars/analysis.groovy
@@ -215,7 +215,7 @@ void gather_timestamps() {
 }
 
 void stash_outcomes(BranchInfo info, String job_name) {
-    def stash_name = job_name + '-outcome'
+    def stash_name = job_name.replace((char) '/', (char) '_') + '-outcome'
     if (findFiles(glob: '*-outcome.csv')) {
         stash(name: stash_name,
               includes: '*-outcome.csv',
@@ -237,8 +237,10 @@ void  analyze_results(Collection<BranchInfo> infos) {
 
         String prefix = infos.size() > 1 ? "$info.branch-" : ''
         String job_name = "${prefix}result-analysis"
-        String outcomes_csv = "${prefix}outcomes.csv"
-        String failures_csv = "${prefix}failures.csv"
+
+        String file_prefix = prefix.replace((char) '/', (char) '_')
+        String outcomes_csv = "${file_prefix}outcomes.csv"
+        String failures_csv = "${file_prefix}failures.csv"
 
         Closure post_checkout = {
             dir('csvs') {
--- a/vars/analysis.groovy
+++ b/vars/analysis.groovy
@@ -258,7 +258,7 @@ void  analyze_results(Collection<BranchInfo> infos) {
             // as test description or test suite was "FAIL".
             if (info.failed_builds) {
                 sh """\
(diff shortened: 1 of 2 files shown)


mpg (approved):
LGTM, thanks for fixing #127 along the way!

gilles-peskine-arm (commented):
Looks good to me on code inspection and with the multibranch runs. Thank you very much for fixing #127!

Before approving I'd just like to see a job with a single branch, and a passing job. I've started https://mbedtls.trustedfirmware.org/job/mbedtls-release-ci-testing/297/ which fulfills both.

bensze01:
@gilles-peskine-arm The CI run you launched has passed.

gilles-peskine-arm (approved):
Both the code and the test jobs look good to me.

gilles-peskine-arm:
The graph with a single branch (e.g. https://mbedtls.trustedfirmware.org/blue/organizations/jenkins/mbedtls-release-ci-testing/detail/mbedtls-release-ci-testing/297/pipeline) looks a bit weird, with the truncated “result-...” and no indication of where “checkout” and “post-execution” belong:

2024-11-18-165707_659x230_scrot

But that's just cosmetic. The important thing there is that the logs of result analysis are now available at the obvious place. Thanks!

bensze01:
The `result-analysis` above the grey line is also supposed to apply to the three steps (`checkout`, `result-analysis` and `post-execution`), but I agree that that is not made as clear by Blue Ocean as it could be.

- above line -> top-level stage the parallel step runs inside
- inline label -> name of the parallel "branch" (this is displayed under the line if there are no internal stages)
- under line -> name of the internal stages within a parallel branch / name of the branch if there are no internal stages.

PR #185, opened 2024-11-20
Add script gathering and reporting health indicators

mpg:
This is a script I've been using to get some stats I see as health indicators for the CI.

### DIFF
--- a/ci_health.py
+++ b/ci_health.py
@@ -1,0 +1,151 @@
+#!/usr/bin/python3
+
+"""
+Gather indicators of CI health using the Jenkins API over the last week.
+
+Currently two indicators are reported:
+    1. Success rate of the nightly jobs. We don't expect "real" failures here,
+    so any failure is likely to be an infra issue or a flaky test.
+    2. Execution time of PR jobs.
+
+Requires python-jenkins.
+(Version 1.4.0-4 from the python3-jenkins package in Ubuntu 24.04 WorksForMe.)
+
+Uses a github token for authentication.
+https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#creating-a-fine-grained-personal-access-token
+As of late 2024, I (mpg) used "classic token" with the following permissions:
+read:discussion, read:enterprise, read:org, read:project, read:user, user:email
+It is likely that a strict subset would work, but I didn't try.
+"""
+
+from statistics import quantiles
+from datetime import datetime, timedelta
+import os
+import sys
+
+import jenkins
+
+# References:
+# 1. https://python-jenkins.readthedocs.io/
+# 2. For each page in the Jenkins web UI, you can append 'api' to the URL (or
+# click the "REST API" link on the bottom left of the page) and then click
+# "Python" on the resulting page to get a preview of what the corresponding
+# API call will return.
+#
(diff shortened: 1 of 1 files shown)


ci_health.py
```diff
@@ -0,0 +1,151 @@
...
+    nb_runs = nb_good + nb_bad
+    success_percent = int(nb_good / nb_runs * 100)
+    print(f"Nightly success rate: {success_percent}% (out of {nb_runs})")
+
+
+def main():
+    """Gather and print out all health indicators."""
+    try:
+        gh_username = os.environ["GITHUB_USERNAME"]
+        gh_token = os.environ["GITHUB_API_TOKEN"]
```
gilles-peskine-arm, line 126: Nice-to-have: I'd like us to standardize on using the GitHub token from `gh` (the official GitHub CLI) when it's available.

I have this code snippet in one of my scripts, which would need to be adapted because here you need the username as well.
```
def try_get_gh_auth_token() -> str:
    """Get the default authentication token from gh (the official GitHub client).

    Return an empty string if there is no such token or if gh is not available.
    """
    # TODO: allow specifying an alternative host name and user name
    try:
        output = subprocess.check_output(['gh', 'auth', 'token'],
                                         stderr=subprocess.DEVNULL)
        return output.strip().decode('ascii')
    except subprocess.CalledProcessError:
        return ''
    except FileNotFoundError:
        return ''
```

ci_health.py
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/python3
+
+"""
+Gather indicators of CI health using the Jenkins API over the last week.
+
+Currently two indicators are reported:
+    1. Success rate of the nightly jobs. We don't expect "real" failures here,
+    so any failure is likely to be an infra issue or a flaky test.
+    2. Execution time of PR jobs.
```
gilles-peskine-arm, line 9: Nice-to-have: an indicator for jobs that fail without a reported cause in the failure list. The failure list is an artifact called `failures.csv` or `failures.csv.xz`. “Without a reported cause” means: `failures.csv.xz` doesn't exist, and (`failures.csv` doesn't exist or `failures.csv` has size 0).

Even better, exclude jobs where the sole failure is that outcome analysis is unhappy.

With https://github.com/Mbed-TLS/mbedtls/pull/9286, which adds an outcome line for running each component, this would count jobs that fail solely due to infrastructure problems (e.g. timeout, network glitches), as well as jobs that fail in outcome analysis. Thus this indicator could become a proxy for jobs that fail solely due to infrastructure problems.

I would ideally like to have an indicator that detects all infrastructure problems, but that seems hard.

ci_health.py
```diff
@@ -0,0 +1,151 @@
...
+# Apparently we're expected to guess by looking at examples?
+#
+# A note about multibranch jobs: there is an extra level of indirection here
+# compared to basic jobs, in the that multibranch job will first give you a
+# list of "sub-jobs", and the builds are associated to those sub-jobs, no the
+# top-level multibranch job.
+
+JENKINS_SERVERS = {
+    "Open": "https://mbedtls.trustedfirmware.org/",
+    "Internal": "https://jenkins-mbedtls.oss.arm.com/",
```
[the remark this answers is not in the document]
mpg: Yep, had that in my local branch but forgot to push until now.

PR #187, opened 2024-12-10
Make readme work

mpg:
Before this PR, if one tries to follow this repo's top-level Readme, things will fail in two places:

1. `scripts/min_requirement.py --user` fails with permission issues. This has been the case for a long time. Fixed by changing `run.sh`.
2. `docker pull` fails. This is new since the tags now include the architecture. Fixed by updating the Readme.

### DIFF
--- a/resources/docker_files/list-docker-image-tags.sh
+++ b/resources/docker_files/list-docker-image-tags.sh
@@ -10,7 +10,11 @@ list () {
     dir="${1%/Dockerfile}"
     dir="${dir%/}"
     hash="$(git hash-object "$dir/Dockerfile")"
-    echo "$(basename -- "$dir")-$hash"
+    base="$(basename -- "$dir")"
+    echo "$base-$hash-amd64"
+    if [ "$base" != "arm-compilers" ]; then
+        echo "$base-$hash-arm64"
+    fi
 }
 
 if [ $# -eq 0 ]; then
--- a/resources/docker_files/run.sh
+++ b/resources/docker_files/run.sh
@@ -56,5 +56,5 @@ echo "  Mounting $SSH_CFG_PATH --> /home/user/.ssh"
 echo "  Mounting $MOUNT_DIR --> /var/lib/ws"
 echo "****************************************************"
 
-sudo docker run --network=host --rm -i -t -u $USR_ID:$USR_GRP -w /var/lib/ws -v $MOUNT_DIR:/var/lib/ws -v $SSH_CFG_PATH:/home/user/.ssh --cap-add SYS_PTRACE ${IMAGE}
+sudo docker run --network=host --rm -i -t -u $USR_ID:$USR_GRP -w /var/lib/ws -e HOME=/var/lib/ws -v $MOUNT_DIR:/var/lib/ws -v $SSH_CFG_PATH:/home/user/.ssh --cap-add SYS_PTRACE ${IMAGE}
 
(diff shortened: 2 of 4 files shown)


Harry-Ramsey (dismissed):
LGTM, I have tested this locally and can confirm it fixes the issue I was experiencing.

resources/docker_files/run.sh
```diff
@@ -56,5 +56,5 @@ echo "  Mounting $SSH_CFG_PATH --> /home/user/.ssh"
 echo "  Mounting $MOUNT_DIR --> /var/lib/ws"
 echo "****************************************************"
 
-sudo docker run --network=host --rm -i -t -u $USR_ID:$USR_GRP -w /var/lib/ws -v $MOUNT_DIR:/var/lib/ws -v $SSH_CFG_PATH:/home/user/.ssh --cap-add SYS_PTRACE ${IMAGE}
+sudo docker run --network=host --rm -i -t -u $USR_ID:$USR_GRP -w /var/lib/ws -e HOME=/var/lib/ws -v $MOUNT_DIR:/var/lib/ws -v $SSH_CFG_PATH:/home/user/.ssh --cap-add SYS_PTRACE ${IMAGE}
```
gilles-peskine-arm, line 59: I think setting `HOME` makes the mounting of `.ssh` not work anymore, since it won't be on `~/.ssh`.

Although did it ever work? The ssh client seems to care that the current user isn't present in the user database.

Should we remove the `.ssh` mounting?

gilles-peskine-arm (approved):
Thanks for updating the documentation! I suspect this breaks the SSH mounting, but I'm not sure it ever worked (it certainly doesn't seem to work just prior to this patch), so this is preexisting and not really made worse.

resources/docker_files/run.sh
bensze01: Yeah, we should - we don't use it for any components, and even if we did, we should pass an ssh-agent socket into the container instead.

mpg: Ah, good point, I removed the `.ssh` mounting because otherwise the command line is self-contradicting which will only cause confusion in the future.

> we should pass an ssh-agent socket into the container instead

mpg: > we should pass an ssh-agent socket into the container instead.

That would certainly be cleaner but apparently it would still run into the issue that the user is not present in `/etc/passwd` and `ssh` is not happy about that: https://gist.github.com/d11wtq/8699521?permalink_comment_id=1904338#gistcomment-1904338

gilles-peskine-arm (approved):
LGTM

Harry-Ramsey (approved):
LGTM.

resources/docker_files/list-docker-image-tags.sh
```diff
@@ -10,7 +10,11 @@ list () {
     dir="${1%/Dockerfile}"
     dir="${dir%/}"
     hash="$(git hash-object "$dir/Dockerfile")"
-    echo "$(basename -- "$dir")-$hash"
+    base="$(basename -- "$dir")"
+    echo "$base-$hash-amd64"
+    if [ "$base" != "arm-compilers" ]; then
+        echo "$base-$hash-arm64"
+    fi
```
gilles-peskine-arm, line 17: Why list both architectures, rather than the native one? [This broke `build.sh`](https://github.com/Mbed-TLS/mbedtls-test/issues/204).

mpg: Most probably because I just wasn't aware how this script would be used (I wasn't even aware it was used by other scripts, as opposed to be purely informational for human consumption). So I assumed we just wanted to list all the tags. So, feel very free to change the behaviour, their was no deep reasoning behind my change.

PR #192, opened 2025-02-21
Delete some sample programs from test

bjwtaylor:
Delete the following programs from test:

pkey/rsa_genkey.c
pkey/pk_decrypt.c
pkey/dh_genprime.c
pkey/rsa_verify.c
pkey/mpi_demo.c
pkey/rsa_decrypt.c
pkey/key_app.c
pkey/dh_server.c
pkey/ecdh_curve25519.c
pkey/pk_encrypt.c
pkey/rsa_sign.c
pkey/key_app_writer.c
pkey/dh_client.c
pkey/ecdsa.c
pkey/rsa_encrypt.c
test/selftest.c
wince_main.c
aes/crypt_and_hash.c
random/gen_random_ctr_drbg.c
random/gen_entropy.c
hash/md_hmac_demo.c
hash/hello.c
hash/generic_sum.c
cipher/cipher_aead_demo.c

### DIFF
--- a/resources/windows/windows_testing.py
+++ b/resources/windows/windows_testing.py
@@ -50,9 +50,7 @@ class VStestrun(object):
         self.run_failed = False
         self.results = OrderedDict([
             ("shipped build", "Not Run"),
-            ("shipped selftest", "Not Run"),
             ("cmake build", "Not Run"),
-            ("cmake selftest", "Not Run"),
             ("cmake test suites", "Not Run"),
         ])
 
@@ -160,11 +158,9 @@ class MbedWindowsTesting(object):
             "Build succeeded.", "\d+ Warning\(s\)", "0 Error\(s\)"
         ]
         self.visual_studio_build_zero_warnings_string = "0 Warning(s)"
-        self.selftest_success_pattern = "\[ All tests (PASS|passed) \]"
         self.test_suites_success_pattern = "100% tests passed, 0 tests failed"
         self.mingw_success_pattern = "PASSED \(\d+ suites, \d+ tests run\)"
         self.config_py_location = os.path.join("scripts", "config.py")
-        self.selftest_exe = "selftest.exe"
         self.mingw_command = "mingw32-make"
         self.git_command = "git"
         self.python_command = "python"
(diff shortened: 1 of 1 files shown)


mpg (approved):
LGTM, thanks! (Based on reviewing the diff, and the test run.)

bensze01:
This will also stop running `selftest.exe` on the LTS branches as well, right?
@mpg @gilles-peskine-arm Is this an expected / desired outcome?

resources/windows/windows_testing.py
```diff
@@ -160,11 +158,9 @@ def __init__(self,
             "Build succeeded.", "\d+ Warning\(s\)", "0 Error\(s\)"
         ]
         self.visual_studio_build_zero_warnings_string = "0 Warning(s)"
-        self.selftest_success_pattern = "\[ All tests (PASS|passed) \]"
```
gilles-peskine-arm, line 163: The test scripts in this repository need to work with all supported branches: `mbedtls-2.28` (for a few more weeks), `mbedtls-3.6` (for a couple more years), and `development`. (It should also work with older versions of these branches, for work in progress, going at least back to the last release in each branch in case we need to release an emergency bug fix.)

As long as the `selftest` program exists, I think we really should keep it running on Windows. So as long as Mbed TLS 3.6 is supported, we need to keep the code related to it in the CI scripts.

What we should do is check if the branch that we're testing has a self-test program, and invoke it if present. So detect if `programs/test/selftest.c` is present, and only invoke and report the selftest if the source file is present.

gilles-peskine-arm (changes requested):
As Bence noted, we need to keep running the selftest program in the LTS branches.

resources/windows/windows_testing.py
bjwtaylor: So, happy with that solution. However do we know if we are likely to have lots of scenarios like this where the latest test is not compatible with the releases/LTS code version? If so I've handled this on other projects before by creating a branch of the CI for release and then each version is run against it's relevant branch. It depends how many changes we expect though, if it's just this one it may not be worth it. We also may need to consider the scenario where we add a test specific to a version that's not relevant to latest. If we have to do this, a branch also maybe useful. Also worth putting this on hold for now until we decide what we are doing with the selftest program.

mpg:
> This will also stop running `selftest.exe` on the LTS branches as well, right? @mpg @gilles-peskine-arm Is this an expected / desired outcome?

Nope, I completely missed that, thanks for catching it!

resources/windows/windows_testing.py
gilles-peskine-arm: It's happened before, and we handled that by having the Groovy code detect the presence or content of a file in the branch being tested.

For example, when we moved from C89 to C99, we had to stop testing the `development` branch with an older version of Visual Studio that didn't support C99, and that decision was made in the Groovy code. So we had the Groovy code [check cmake settings in the branch under test](https://github.com/Mbed-TLS/mbedtls-test/commit/9518538efd49c9d8089657e9a3bf0ea1e7e9a4b2#diff-ac212fe0bc0d1e910ac7afcd3a6ca2a808f316870aa1e36ef11eeb7ab88a563bR114) to determine whether C89 or C99 was desired.

bjwtaylor:
This will no longer be required, as we have now decided to not remove selftest. See https://arm-ce.slack.com/archives/C07PFD5GE5C/p1740729895646409 for more detais.

PR #193, opened 2025-02-24
Use a venv for min_requirements.py on Linux and FreeBSD

bensze01:
Use a python virtual environment when installing pip packages via min_requirements.py. This prevents the installed packages from interfering with the host system.

It also works around spurious build failures when installing said packages. Fixes https://github.com/Mbed-TLS/mbedtls-test/issues/194

Test runs:
- development
  - Internal CI
    - [mbed-tls-pr-test-parametrized][1]
  - Open CI
    - [mbed-tls-restricted-pr-test-parametrized][2] 
  
[1]: https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-test-parametrized/448/
[2]: https://mbedtls.trustedfirmware.org/job/mbed-tls-restricted-pr-test-parametrized/96/

### DIFF
--- a/vars/gen_jobs.groovy
+++ b/vars/gen_jobs.groovy
@@ -142,7 +142,9 @@ set -eux
 ulimit -f 20971520
 
 if [ -e scripts/min_requirements.py ]; then
-scripts/min_requirements.py --user ${info.python_requirements_override_file}
+python3 -m venv --system-site-packages --without-pip venv
+export PATH="\$PWD/venv/bin:\$PATH"
+python3 scripts/min_requirements.py ${info.python_requirements_override_file}
 fi
 """ + script_in_docker
                     sh 'chmod +x steps.sh'
@@ -247,7 +249,9 @@ echo >&2 'Note: "clang" will run /usr/bin/clang -Wno-error=c11-extensions'
 
     if (info.has_min_requirements) {
         extra_setup_code += """
-scripts/min_requirements.py --user ${info.python_requirements_override_file}
+python3 -m venv --system-site-packages --without-pip venv
+export PATH="\$PWD/venv/bin:\$PATH"
+scripts/min_requirements.py ${info.python_requirements_override_file}
 """
     }
 
--- a/vars/gen_jobs.groovy
+++ b/vars/gen_jobs.groovy
@@ -251,7 +251,7 @@ echo >&2 'Note: "clang" will run /usr/bin/clang -Wno-error=c11-extensions'
         extra_setup_code += """
 python3 -m venv --system-site-packages --without-pip venv
 export PATH="\$PWD/venv/bin:\$PATH"
-scripts/min_requirements.py ${info.python_requirements_override_file}
+python3 scripts/min_requirements.py ${info.python_requirements_override_file}
 """
     }
 
--- a/resources/docker_files/arm-compilers/Dockerfile
+++ b/resources/docker_files/arm-compilers/Dockerfile
@@ -92,6 +92,8 @@ RUN apt-get update -q && apt-get install -yq \
         locales \
         # to install several Python packages (done by individual jobs)
         python3-pip \
+        # to install several Python packages (done by individual jobs)
+        python3-venv \
     && rm -rf /var/lib/apt/lists/
 
 # Install Python pip packages
--- a/resources/docker_files/ubuntu-18.04/Dockerfile
+++ b/resources/docker_files/ubuntu-18.04/Dockerfile
@@ -83,8 +83,10 @@ RUN apt-get update -q && apt-get install -yq \
         make \
         # to build GnuTLS with locally-compiled nettle
         pkg-config \
-        # to install the preferred version of pylint
+        # to install several Python packages (done by individual jobs)
         python3-pip \
+        # to install several Python packages (done by individual jobs)
+        python3-venv \
         # for Mbed TLS tests
         valgrind \
         # to download things installed from other places
--- a/resources/docker_files/ubuntu-20.04/Dockerfile
+++ b/resources/docker_files/ubuntu-20.04/Dockerfile
@@ -91,6 +91,8 @@ RUN apt-get update -q && apt-get install -yq \
         pkg-config \
         # to install several Python packages (done by individual jobs)
         python3-pip \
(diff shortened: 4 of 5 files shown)


vars/gen_jobs.groovy
```diff
@@ -247,7 +249,9 @@ echo >&2 'Note: "clang" will run /usr/bin/clang -Wno-error=c11-extensions'
 
     if (info.has_min_requirements) {
         extra_setup_code += """
-scripts/min_requirements.py --user ${info.python_requirements_override_file}
+python3 -m venv --system-site-packages --without-pip venv
+export PATH="\$PWD/venv/bin:\$PATH"
+scripts/min_requirements.py ${info.python_requirements_override_file}
```
yanesca, line 254: Why don't we call python3 explicitly here, like we did above?

vars/gen_jobs.groovy
```diff
@@ -142,7 +142,9 @@ set -eux
 ulimit -f 20971520
 
 if [ -e scripts/min_requirements.py ]; then
-scripts/min_requirements.py --user ${info.python_requirements_override_file}
+python3 -m venv --system-site-packages --without-pip venv
```
yanesca, line 145: Why do we need the `--without-pip` option?

yanesca (commented):
I left a few questions inline.

vars/gen_jobs.groovy
```diff
@@ -142,7 +142,9 @@ set -eux
...
+export PATH="\$PWD/venv/bin:\$PATH"
+python3 scripts/min_requirements.py ${info.python_requirements_override_file}
```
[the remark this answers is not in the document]
bensze01: Since we are installing inside a `venv`, we the "system" directory  inside the venv is already user-writable, so we don't need to use the alternate "user" path.

vars/gen_jobs.groovy
bensze01: If we don't supply `--without-pip`, the default version of `pip`, the distro's default version of `pip` will be bootstrapped inside the `venv` via `python3 -m ensurepip`.

Since we also pass `--system-site-packages`, we can still call the version of `pip` we install in the Dockerfiles inside the venv via `venv/bin/python3 -m pip`

yanesca: Oh, never mind, I see it now, when initialising the venv it will get the distro pip even if the global pip has been updated since.

yanesca: I am just worried a bit that `--system-site-packages` breaks the isolation. I feel it would be more robust to query the pip version before creating the environment and then force-reinstalling the same version inside. Do you think that could work? Would that be very complicated?

bensze01: As far as I know, packages installed inside the `venv` should take precedence over system-wide packages, and `pip` should up/downgrade any packages inside the `venv` if the version outside is not compatible with the packages we are installing, so the isolation should be fine.

I've run into some issues with our current set-up on Ubuntu 16.04 (where we install a different version of python manually) when I didn't supply `--without-pip`, so I'd prefer to stick to my original solution for the emergency fix - we can come back around to figure out something more robust afterwards in a follow-up PR in my opinion.

gilles-peskine-arm (commented):
I'm not necessarily opposed to approving this as an emergency fix, because I can see that it makes the job pass. But I don't understand how this fixes the problem, and I don't understand what caused the problem in the first place. When I compare logs from an [old working job](https://mbedtls.trustedfirmware.org/job/mbed-tls-nightly-tests/1431/execution/node/5270/log/?consoleFull), a [failing job](https://mbedtls.trustedfirmware.org/job/mbed-tls-nightly-tests/1431/execution/node/5270/log/?consoleFull), and [this fix](https://mbedtls.trustedfirmware.org/blue/rest/organizations/jenkins/pipelines/mbed-tls-restricted-pr-test-parametrized/runs/96/nodes/743/steps/5759/log/?start=0), I don't understand what changed. In particular, we seem to always be grabbing the same versions of packages and their dependencies, so why did it start failing?

vars/gen_jobs.groovy
```diff
@@ -142,7 +142,9 @@ set -eux
 ulimit -f 20971520
 
 if [ -e scripts/min_requirements.py ]; then
-scripts/min_requirements.py --user ${info.python_requirements_override_file}
+python3 -m venv --system-site-packages --without-pip venv
+export PATH="\$PWD/venv/bin:\$PATH"
```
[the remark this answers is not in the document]
bensze01: The venv's activate script effectively  doesn't do much other than set the `PATH`.

Python figures out if it is being run from within a venv by looking at the location of the interpreter (symlink), and seeing if it is a subdirectory of a valid venv.

vars/gen_jobs.groovy
```diff
@@ -247,7 +249,9 @@ echo >&2 'Note: "clang" will run /usr/bin/clang -Wno-error=c11-extensions'
 
     if (info.has_min_requirements) {
         extra_setup_code += """
-scripts/min_requirements.py --user ${info.python_requirements_override_file}
+python3 -m venv --system-site-packages --without-pip venv
```
gilles-peskine-arm, line 252: Since we're using a virtual environment, why not use it fully? Why would we use this hybrid mode where we still use some system packages and the system pip?

yanesca: I asked the same (https://github.com/Mbed-TLS/mbedtls-test/pull/193#discussion_r1967591214) and we can do that later as an improvement, but Bence prefers to do this quick fix now and potentially improve on it later.

vars/gen_jobs.groovy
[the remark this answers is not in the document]
yanesca: > The venv's activate script effectively doesn't do much other than set the PATH

As far as I could tell, the main feature (beyond setting the path) is to set up the deactivate function to restore the environment and an environment variable that can be used for some convenience features (for example to indicate the virtual environment in your bash prompt).

gilles-peskine-arm (approved):
Approved as an emergency fix, on the condition that we treat https://github.com/Mbed-TLS/mbedtls-test/issues/195 as a high-priority follow-up.

PR #199, opened 2025-03-26
Update OPENSSL_NEXT to what 3.6 and development use

mpg:
3.6 and development want `OPENSSL_NEXT` to be 3.1.2.

In `all.sh` this is forced, but not in `basic-build-test.sh` (used by the coverage part of the release job), or when invoking `ssl-opt.sh` directly (which does not happen on the CI but might when using the Docker images for debugging).

As a result, `basic-build-test.sh` has been failing in the 3.6 release CI (because the new tests crucially rely on `OPENSSL_NEXT` being 3.x).

This can be fixed by updating `OPENSSL_NEXT` in the docker images. We couldn't do it previously because 2.28 was still relying on it being 1.1.1a, but now that 2.28 is EOL we're finally free.

(Note that in the long run, the plan is to no longer set `OPENSSL_NEXT` or `OPENSSL_LEGACY` in the Docker files. But first we need to update all script to use named version variables like `OPENSSL_3` - and then make a release - and only then we can remove the variables from the Dockerfiles.)

Test builds:
- development:  https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/793
- 3.6: https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/794/

### DIFF
--- a/resources/docker_files/ubuntu-22.04/Dockerfile
+++ b/resources/docker_files/ubuntu-22.04/Dockerfile
@@ -322,6 +322,9 @@ RUN python3 -m pip config set global.progress_bar off && \
     python3 -m pip install setuptools --upgrade && \
     true
 
+#Upgrade pip to the latest version
+RUN  python3 -m pip install --upgrade pip
+
 # Make sure we have a UTF-8 locale
 RUN locale && \
     locale-gen "en_US.UTF-8" && \
--- a/resources/docker_files/ubuntu-22.04/Dockerfile
+++ b/resources/docker_files/ubuntu-22.04/Dockerfile
@@ -227,8 +227,6 @@ RUN wget -q https://www.openssl.org/source/openssl-1.1.1a.tar.gz && \
 ENV OPENSSL_1_1_1=/usr/local/openssl-1.1.1a/bin/openssl
 ENV OPENSSL_1_1=/usr/local/openssl-1.1.1a/bin/openssl
 ENV OPENSSL_1=/usr/local/openssl-1.1.1a/bin/openssl
-# Up to at least 2023, the test scripts expect that $OPENSSL_NEXT is 1.1.1a.
-ENV OPENSSL_NEXT=/usr/local/openssl-1.1.1a/bin/openssl
 
 RUN wget -q https://www.openssl.org/source/openssl-3.0.10.tar.gz && \
     tar -zxf openssl-3.0.10.tar.gz && cd openssl-3.0.10 && \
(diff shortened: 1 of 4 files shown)


davidhorstmann-arm (approved):
LGTM, thanks!

minosgalanakis (approved):
LGTM

minosgalanakis:
+1 on test parameters.

mpg:
@minosgalanakis In order to test your new commit you need to also push to dev/mpg/fix-openssl-next in this repo. The CI is not picking code from mbedtls-test forks.

Arguably it was my mistake to create this PR from a fork, I should have created it from a branch in this repo - this would have ensured that what we test on the CI is the same as what we see in the PR. But I had forgotten about that restriction, so here we are: each change needs to be push both to mpg:fix-openssl-next (so that it shows in the PR) _and_ mbedtls-test:dev/mpg/fix-openssl-next (so that it can be picked by the CI).

I that's too annoying, we can close this PR (github doesn't allow changing the source branch on an existing PR) and re-open one with a saner branching scheme.

mpg:
I've just renamed my remote from `mpg` to `mpg-careful-the-ci-cant-pick-code-from-forks` so hopefully next time I'll remember that PRs to mbedtls-test that need CI test runs should not be created from my fork :)

minosgalanakis:
The test jobs have passed [includig the code-coverage](https://jenkins-mbedtls.oss.arm.com/blue/organizations/jenkins/mbedtls-release-ci-testing/detail/mbedtls-release-ci-testing/794/pipeline/1644) step.

This is ready for review

resources/docker_files/ubuntu-22.04/Dockerfile
```diff
@@ -316,6 +322,9 @@ RUN python3 -m pip config set global.progress_bar off && \
     python3 -m pip install setuptools --upgrade && \
     true
 
+#Upgrade pip to the latest version
+RUN  python3 -m pip install --upgrade pip
```
mpg, line 326: I think ideally we should merge this with the previous command:
```
RUN python3 -m pip config set global.progress_bar off && \
    python3 -m pip install --upgrade pip && \
    python3 -m pip install --upgrade setuptools && \
    true
```
(It makes more sense to update pip before using it to update setuptools, and to group all python update together.)

But that's minor and I don't want to hold this PR back.

mpg (commented):
[Rewiewing the commit from Minos, not mine]

Looks good to me on code review. (Just one comment but it's minor and can be done later.)

Regarding the CI, it ran on a different commit, https://github.com/Mbed-TLS/mbedtls-test/commit/3060c99d275cff3f1aa6df5071b28491f254deda from the commit in this PR https://github.com/Mbed-TLS/mbedtls-test/pull/199/commits/a800ada9a5360d29e6a8bce8607761b7ab7f5a62 - that is, I manually check that both commits have the same content, but they have different hashes. That's not ideal, and usually I would have requested a re-run with the same commit, but I don't want to hold this PR back so I'm OK approving based on manually checking that both commits have identical contents to help move this PR forward.

resources/docker_files/ubuntu-22.04/Dockerfile
minosgalanakis: I did try the following in the reverse order and it failed. 

~~~~~
RUN python3 -m pip config set global.progress_bar off && \
    python3 -m pip install --upgrade setuptools && \
    python3 -m pip install --upgrade pip && \
    true
~~~~~

Ideally we need to  figure out the versions of pip and setup tools that have addressed the bug and then pin it there, but as you said this could be another patch. ( I am more than happy to do i after this one )

minosgalanakis (approved):
Review of commit [e7a22c5](https://github.com/Mbed-TLS/mbedtls-test/pull/199/commits/e7a22c5bc8b781a707cb06fe9f97c125dcf275ef)

I have locally built all docker images, and verified that OPENSSL_NEXT is being set properly.  The 2204 image is failing because of an unrelated bug.

davidhorstmann-arm (approved):
LGTM, thanks!

PR #201, opened 2025-04-22
Checkout from GitHub only once per branch, stash and then just unstash

### LINKED ISSUE #197: Use a consistent commit from consuming branches
There is a race condition on the CI for https://github.com/Mbed-TLS/mbedtls-framework and https://github.com/Mbed-TLS/TF-PSA-Crypto. They perform some testing with consuming branches from https://github.com/Mbed-TLS/mbedtls. It uses the head of the consuming branch. If the consuming branch moves during the execution of the script, different parts of the job may test against different commits. This can result in spurious failures (or potentially spurious passes, but I don't think that's likely).

Example: in [a merge queue job on TF-PSA-Crypto](https://mbedtls.trustedfirmware.org/blue/organizations/jenkins/mbed-tls-tf-psa-crypto-multibranch/detail/gh-readonly-queue%2Fdevelopment%2Fpr-195-399c5f9e1d71cb177eb0c16cb934755b409abe23/1/pipeline/16541/), most of the CI run used the mbedtls development branch at [eb20c1f12d54aefba8a22c293721900b1d63a1cb](https://github.com/Mbed-TLS/mbedtls/tree/eb20c1f12d54aefba8a22c293721900b1d63a1cb), but the result-analysis job used a more recent commit [50432e42c88991743d4b4a9bb732d5313fe026e8](https://github.com/Mbed-TLS/mbedtls/tree/50432e42c88991743d4b4a9bb732d5313fe026e8). The more recent commit had additional test cases, and those test cases were never executed (because they weren't merged yet when the `all.sh` components ran), but they were picked up by result analysis since they had been merged by that time.

ronald-cron-arm:
Fixes #197 and reduce significantly the number of requests to GitHub.
- OpenCI
  - TF-PSA-Crypto
    - [x] [PR-head](https://mbedtls.trustedfirmware.org/job/mbed-tls-tf-psa-crypto-restricted-ci-testing/view/change-requests/job/PR-3-head/22/)
  - Mbed-TLS
    - development
      - [x] [PR-head](https://mbedtls.trustedfirmware.org/job/mbed-tls-restricted-pr-ci-testing/view/change-requests/job/PR-906-head/36/)
      - [x] [PR-merge](https://mbedtls.trustedfirmware.org/job/mbed-tls-restricted-pr-ci-testing/view/change-requests/job/PR-906-merge/56/)
    - 3.6
      - [x] [PR-head](https://mbedtls.trustedfirmware.org/job/mbed-tls-restricted-pr-ci-testing/view/change-requests/job/PR-1233-head/11/)
      - [x] [PR-merge](https://mbedtls.trustedfirmware.org/job/mbed-tls-restricted-pr-ci-testing/view/change-requests/job/PR-1233-merge/17)
    - release
      - [x] [development + 3.6](https://mbedtls.trustedfirmware.org/job/mbedtls-release-ci-testing/368/)

- Internal CI
  - TF-PSA-Crypto
    - [x] [PR-head](https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-tf-psa-crypto-restricted-ci-testing/view/change-requests/job/PR-3-head/14/)
  - Mbed-TLS
    - development
      - [x] [PR-head](https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-ci-testing/view/change-requests/job/PR-906-head/135/)
      - [x] [PR-merge](https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-ci-testing/view/change-requests/job/PR-906-merge/162/)
    - 3.6
      - [x] [PR-head](https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-ci-testing/view/change-requests/job/PR-1233-head/14/)
      - [x] [PR-merge](https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-ci-testing/view/change-requests/job/PR-1233-merge/13/)
    - release
      - [x] [development + 3.6](https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/805/)

### DIFF
--- a/vars/checkout_repo.groovy
+++ b/vars/checkout_repo.groovy
@@ -150,15 +150,23 @@ git submodule foreach --recursive git config url.git@github.com:.insteadOf https
 }
 
 void checkout_repo(BranchInfo info) {
-    switch(info.repo) {
-        case 'tls':
-            checkout_tls_repo(info)
-            break
-        case 'tf-psa-crypto':
-            checkout_tf_psa_crypto_repo(info)
-            break
-        default:
-            error("Invalid repo: $info.repo")
+    if (!info.stash) {
+        switch (info.repo) {
+            case 'tls':
+                checkout_tls_repo(info)
+                break
+            case 'tf-psa-crypto':
+                checkout_tf_psa_crypto_repo(info)
+                break
+            default:
+                error("Invalid repo: ${info.repo}")
+        }
+
+        def stashName = "${info.job_prefix}stash"
+        stash name: stashName, includes: '**/*', useDefaultExcludes: false
+        info.stash = stashName
+    } else {
+        unstash info.stash
     }
 }
 
--- a/vars/checkout_repo.groovy
+++ b/vars/checkout_repo.groovy
@@ -163,7 +163,7 @@ void checkout_repo(BranchInfo info) {
                     error("Invalid repo: ${info.repo}")
             }
 
-            def stashName = "${info.job_prefix}stash"
+            def stashName = "${info.prefix}stash"
             stash name: stashName, includes: '**/*', useDefaultExcludes: false
             info.stash = stashName
         } else {
--- a/vars/checkout_repo.groovy
+++ b/vars/checkout_repo.groovy
@@ -150,23 +150,32 @@ git submodule foreach --recursive git config url.git@github.com:.insteadOf https
 }
 
 void checkout_repo(BranchInfo info) {
+    def stashName = "${info.prefix}stash"
+    def needUnstash = true
+
     try {
         if (!info.stash) {
-            switch (info.repo) {
-                case 'tls':
-                    checkout_tls_repo(info)
-                    break
-                case 'tf-psa-crypto':
-                    checkout_tf_psa_crypto_repo(info)
-                    break
-                default:
-                    error("Invalid repo: ${info.repo}")
-            }
+            lock(resource: "checkout_repo-${stashName}", inversePrecedence: true) {
+                if (!info.stash) {
+                    switch (info.repo) {
+                        case 'tls':
+                            checkout_tls_repo(info)
+                            break
+                        case 'tf-psa-crypto':
+                            checkout_tf_psa_crypto_repo(info)
+                            break
+                        default:
+                            error("Invalid repo: ${info.repo}")
+                    }
 
-            def stashName = "${info.prefix}stash"
-            stash name: stashName, includes: '**/*', useDefaultExcludes: false
-            info.stash = stashName
-        } else {
+                    stash name: stashName, includes: '**/*', useDefaultExcludes: false
+                    info.stash = stashName
+                    needUnstash = false
+                }
+            }
+        }
+
+        if (needUnstash) {
             unstash info.stash
         }
     } catch (exception) {
--- a/vars/checkout_repo.groovy
+++ b/vars/checkout_repo.groovy
@@ -150,23 +150,28 @@ git submodule foreach --recursive git config url.git@github.com:.insteadOf https
 }
 
 void checkout_repo(BranchInfo info) {
-    if (!info.stash) {
-        switch (info.repo) {
-            case 'tls':
-                checkout_tls_repo(info)
-                break
-            case 'tf-psa-crypto':
-                checkout_tf_psa_crypto_repo(info)
-                break
-            default:
-                error("Invalid repo: ${info.repo}")
-        }
+    try {
+        if (!info.stash) {
+            switch (info.repo) {
+                case 'tls':
+                    checkout_tls_repo(info)
+                    break
+                case 'tf-psa-crypto':
+                    checkout_tf_psa_crypto_repo(info)
+                    break
+                default:
+                    error("Invalid repo: ${info.repo}")
+            }
 
-        def stashName = "${info.job_prefix}stash"
-        stash name: stashName, includes: '**/*', useDefaultExcludes: false
-        info.stash = stashName
-    } else {
-        unstash info.stash
+            def stashName = "${info.job_prefix}stash"
+            stash name: stashName, includes: '**/*', useDefaultExcludes: false
+            info.stash = stashName
+        } else {
+            unstash info.stash
+        }
+    } catch (exception) {
+        echo "Caught: ${common.stack_trace_to_string(exception)}"
+        throw exception
     }
 }
 
--- a/vars/checkout_repo.groovy
+++ b/vars/checkout_repo.groovy
@@ -155,7 +155,7 @@ void checkout_repo(BranchInfo info) {
 
     try {
         if (!info.stash) {
-            lock(resource: "checkout_repo-${stashName}", inversePrecedence: true) {
+            lock(resource: "stash-lock/${env.BUILD_TAG}-${stashName}") {
                 if (!info.stash) {
                     switch (info.repo) {
                         case 'tls':
--- a/vars/common.groovy
+++ b/vars/common.groovy
@@ -340,14 +340,10 @@ List<BranchInfo> get_branch_information(Collection<String> tls_branches, Collect
     platform_jobs.failFast = true
     def platform_results = (Map<String, Map<String, String>>) parallel(platform_jobs)
 
-    def results = [:]
-    results.putAll(all_results)
-    results.putAll(platform_results)
-
     infos.each { BranchInfo info ->
-        info.all_sh_components = results[info.job_prefix + 'all-platforms']
+        info.all_sh_components = all_results[info.job_prefix + 'all-platforms']
         linux_platforms.reverseEach { platform ->
-            info.all_sh_components << results[info.job_prefix + platform]
+            info.all_sh_components << platform_results[info.job_prefix + platform]
         }
 
         if (env.JOB_TYPE == 'PR') {
--- a/vars/common.groovy
+++ b/vars/common.groovy
@@ -226,8 +226,8 @@ docker run -u \$(id -u):\$(id -g) -e MAKEFLAGS -e VERBOSE_LOGS $env_args --rm --
  * In particular, get components of all.sh for Linux platforms. */
 List<BranchInfo> get_branch_information(Collection<String> tls_branches, Collection<String> tf_psa_crypto_branches) {
     List<BranchInfo> infos = []
-    Map<String, Object> all_jobs = [:]
-    Map<String, Object> platform_jobs = [:]
+    Map<String, Object> list_all_sh_components_jobs = [:]
+    Map<String, Object> list_platform_components_jobs = [:]
 
     Map<String, Collection<String>> repos = ['tls': tls_branches, 'tf-psa-crypto': tf_psa_crypto_branches]
     // Filter out repos with no branches
@@ -247,7 +247,7 @@ List<BranchInfo> get_branch_information(Collection<String> tls_branches, Collect
                 info.prefix += "$branch-"
             }
 
-            all_jobs << gen_jobs.job(info.prefix + 'all-platforms') {
+            list_all_sh_components_jobs << gen_jobs.job(info.prefix + 'all-platforms') {
                 node('container-host') {
                     try {
                         // Log the environment for debugging purposes
@@ -297,7 +297,7 @@ List<BranchInfo> get_branch_information(Collection<String> tls_branches, Collect
             }
 
             linux_platforms.each { platform ->
-                platform_jobs << gen_jobs.job(info.prefix + platform) {
+                list_platform_components_jobs << gen_jobs.job(info.prefix + platform) {
                     node(gen_jobs.node_label_for_platform(platform)) {
                         try {
                             dir('src') {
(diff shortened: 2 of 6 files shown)


vars/common.groovy
```diff
@@ -333,8 +334,15 @@ List<BranchInfo> get_branch_information(Collection<String> tls_branches, Collect
...
-    def results = (Map<String, Map<String, String>>) parallel(jobs)
+    all_jobs.failFast = true
+    def all_results = (Map<String, Map<String, String>>) parallel(all_jobs)
+
+    platform_jobs.failFast = true
+    def platform_results = (Map<String, Map<String, String>>) parallel(platform_jobs)
+
+    def results = [:]
+    results.putAll(all_results)
+    results.putAll(platform_results)
```
bensze01, line 345: We can simplify this slightly
```diff
diff --git a/vars/common.groovy b/vars/common.groovy
index 268bfb6..2e5682c 100644
--- a/vars/common.groovy
+++ b/vars/common.groovy
@@ -340,14 +340,10 @@ List<BranchInfo> get_branch_information(Collection<String> tls_branches, Collect
     platform_jobs.failFast = true
     def platform_results = (Map<String, Map<String, String>>) parallel(platform_jobs)

-    def results = [:]
-    results.putAll(all_results)
-    results.putAll(platform_results)
-
     infos.each { BranchInfo info ->
-        info.all_sh_components = results[info.job_prefix + 'all-platforms']
+        info.all_sh_components = all_results[info.job_prefix + 'all-platforms']
         linux_platforms.reverseEach { platform ->
-            info.all_sh_components << results[info.job_prefix + platform]
+            info.all_sh_components << platform_results[info.job_prefix + platform]
         }

         if (env.JOB_TYPE == 'PR') {
```

bensze01 (dismissed):
One small nit, otherwise looks good to me.

vars/common.groovy
ronald-cron-arm: Indeed, thanks.

vars/checkout_repo.groovy
```diff
@@ -150,15 +150,28 @@ git submodule foreach --recursive git config url.git@github.com:.insteadOf https
...
+                    break
+                case 'tf-psa-crypto':
+                    checkout_tf_psa_crypto_repo(info)
+                    break
+                default:
+                    error("Invalid repo: ${info.repo}")
+            }
+
+            def stashName = "${info.job_prefix}stash"
+            stash name: stashName, includes: '**/*', useDefaultExcludes: false
```
gilles-peskine-arm, line 167: If `checkout_repo` is invoked with the same `info.job_prefix` value in parallel, there is a race condition: whoever reaches the `stash` call first will error out because the stash name already exists.

This doesn't seem to be a problem given the way we currently use `checkout_repo`, but it looks fragile to me. For example, maybe it's a problem only if we run certain subsets of the jobs? Please at least document it, though I'd feel more comfortable with a lock.

vars/checkout_repo.groovy
```diff
@@ -150,15 +150,28 @@ git submodule foreach --recursive git config url.git@github.com:.insteadOf https
...
+            info.stash = stashName
+        } else {
+            unstash info.stash
```
gilles-peskine-arm, line 170: The [Jenkins documentation](https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#stash-stash-some-files-to-be-used-later-in-the-build) says:

> Note that the stash and unstash steps are designed for use with small files. For large data transfers, use the External Workspace Manager plugin, or use an external repository manager such as Nexus or Artifactory. This is because stashed files are archived in a compressed TAR, and with large files this demands considerable resources on the controller, particularly CPU time. There's not a hard stash size limit, but between 5-100 MB you should probably consider alternatives.

An mbedtls-3.6 source tree with the framework is about 27MB. Should we fear performance problems?

Note that we already use stashes for outcomes, for which I've also wondered this and never gotten an answer. Those are mostly about 8MB apiece, with a few larger ones. So if the size is a problem, it's just making an existing problem worse.

vars/common.groovy
```diff
@@ -246,7 +247,7 @@ List<BranchInfo> get_branch_information(Collection<String> tls_branches, Collect
                 info.job_prefix += "$branch-"
             }
 
-            jobs << gen_jobs.job(info.job_prefix + 'all-platforms') {
+            all_jobs << gen_jobs.job(info.job_prefix + 'all-platforms') {
```
gilles-peskine-arm, line 250: I don't understand the variable names `all_jobs` and `platform_jobs` in this function. I think how `platform_jobs` has one entry per platform, so that sort of makes sense. But `all_jobs` is neither “all the jobs” in some sense, nor “jobs related to `all.sh`”, and I'm confused.

gilles-peskine-arm (changes requested):
Looks good to me functionally (pending CI results for the latest commit), but I have a couple of concerns about performance and concurrency.

vars/checkout_repo.groovy
ronald-cron-arm: By construction of `info.job_prefix` it is not possible to have different `info` objects with the same `job_prefix`. Thus there is no need for a lock as the code stands. But if you still prefer I can add a lock. Do we have a example of that?

gilles-peskine-arm: > By construction of info.job_prefix it is not possible to have different info objects with the same job_prefix

This is not at all obvious to me, and I can't see it documented anywhere. If it's an expected invariant, I'd like it to be documented. Even if it's the case now, given the lack of documentation, it may not be the case after we add some feature or other.

> But if you still prefer I can add a lock. Do we have a example of that?

Going by example in `gen_jobs.groovy`, I guess something like
```
lock("checkout-repo/${info.job_prefix}") { … }
```

vars/checkout_repo.groovy
ronald-cron-arm: It seems you are right we are likely to overload the controller. @bensze01 do you confirm? It seems an alternative is to use an "Artifact Manager (e.g., S3, Artifactory)". @bensze01 do you know if we can use that in the internal and Open CI?

vars/checkout_repo.groovy
ronald-cron-arm: I've tried to improve the code and the documentation rather than adding a lock.

vars/common.groovy
ronald-cron-arm: I've tried to improve the variable names.

vars/checkout_repo.groovy
ronald-cron-arm: After discussions with Bence we think it is worth giving it a try. The controller does not seem a bottleneck currently and the size of a stash (~50M) is not that big. If it does not scale, it is easy to revert and we can then try to use an "S3 bucket" to handle the unstash traffic in place of the controller.

vars/checkout_repo.groovy
gilles-peskine-arm: Thanks for improving the documentation and cleaning up the code! But it's still not clear to me why `checkout_repo` would never be called concurrently on the same repo+branch. Actually, that does happen, doesn't it? What makes it work is that the _first_ call for a given repo+branch is always done during some setup phase that is unique and must complete before the repo+branch is used again. That property seems pretty fragile (e.g. we might break it if we add a Python setup phase that's independent of the `all.sh` gathering phase), and it's undocumented.

ronald-cron-arm: That's a good point, thanks. I will improve that.

gilles-peskine-arm (commented):
LGTM on code reading at 479ce7e3ac28464119db6839600a53c755f3c0eb except that I'd like documentation or code to address https://github.com/Mbed-TLS/mbedtls-test/pull/201#discussion_r2063207288.

vars/gen_jobs.groovy
```diff
@@ -458,10 +458,28 @@ def gen_abi_api_checking_job(BranchInfo info, String platform) {
...
+                    https://* )
+                        ;;
+                    * )
+                    # SCP-like git URLs not handled
+                    echo "Unknown URL type: $origin_url"
+                    ;;
+                esac
+
+                git fetch --depth 1 origin ${CHANGE_TARGET}
+            '''
```
bensze01: This can be simplified down to a single-line:
```suggestion
sh "GIT_SSH_COMMAND='ssh -o StrictHostKeyChecking=accept‐new' git fetch --depth 1 origin ${CHANGE_TARGET}"
```

vars/checkout_repo.groovy
```diff
@@ -150,23 +150,32 @@ git submodule foreach --recursive git config url.git@github.com:.insteadOf https
...
-            switch (info.repo) {
-                case 'tls':
-                    checkout_tls_repo(info)
-                    break
-                case 'tf-psa-crypto':
-                    checkout_tf_psa_crypto_repo(info)
-                    break
-                default:
-                    error("Invalid repo: ${info.repo}")
+            lock(resource: "checkout_repo-${stashName}", inversePrecedence: true) {
```
bensze01, line 158: Use local per-branch locks instead of global ones
```suggestion
            synchronized (info) {
```

bensze01 (changes requested):
I've requested two changes simplifying the code and reducing resource contention.
Otherwise looks good to me.

vars/gen_jobs.groovy
ronald-cron-arm: This is indeed much shorter but our CI does not support the "accept-new" option. Otherwise, in terms of security it is not as good as with this approach we trust blindly the first connection. Actually I'd say there is no authentication at all but I may miss something.

vars/checkout_repo.groovy
ronald-cron-arm: Doing some researches it seems that this does not work in the context of Jenkins pipelines as Jenkins Pipelines are not multi-threaded in the JVM sense.

bensze01: Are you sure?
Jenkins readily offers thread dumps in the java sense for parallel stages.
I was under the impression that Jenkins pipelines basically have a Global Interpreter Lock like Python, which they release for long-running steps like `checkout` or `sh`.

So `synchronized` should work just fine.

vars/gen_jobs.groovy
```diff
@@ -427,50 +427,76 @@ def gen_windows_testing_job(BranchInfo info, String toolchain, String label_pref
...
+            sshagent([env.GIT_CREDENTIALS_ID]) {
+                sh '''#!/bin/sh
+                    origin_url=$(git remote get-url origin)
+                    echo "origin remote URL: $origin_url"
+                    case "$origin_url" in
+                        ssh://* )
+                            domain=$(printf "%s" "$origin_url" | sed -n 's#ssh://[^@]*@\\([^/:]*\\).*#\\1#p')
+                            echo "origin remote domain: $domain"
+                            mkdir -p ~/.ssh
+                            ssh-keyscan "$domain" >> ~/.ssh/known_hosts
```
gilles-peskine-arm, line 474: This makes me uncomfortable. We're potentially enabling man-in-the-middle attacks. Why do we need this? It doesn't seem related to the rest of the PR.

gilles-peskine-arm (changes requested):
LGTM on code reading except I don't understand the SSH thing.

vars/gen_jobs.groovy
bensze01: This is what the OpenCI is configured to do at the moment when we do checkouts.
But this change is applied on a per-node basis, and we don't call `checkout` on all nodes anymore, leading to intermittent failures (since there is no interactive user to type `yes` when ssh sees a new server)
Our alternative solution is to ask and wait for Linaro to hard-code Github's public key on our AMI machine images.

gilles-peskine-arm: Ok, thanks for the information, but why wasn't it a problem before? It's not that we were updating `known_hosts` only on the master before and now we do it on all the nodes. If the master has reliable information and the nodes don't, then we should copy the information from the master rather than obtain it in an unreliable way.

bensze01: `checkout` uses `ssh` checkouts directly on the nodes.
We had issues before when Github updated their `ssh` public keys (due to a breach), and after that Linaro configured Jenkins to `accept new keys for unknown hosts`

This means that when calling `checkout` jenkins is configured to accept and add the public key of any new/unkown host to `known_hosts` on the node the checkout is running on.

But this file gets reset every time the node spins down due to inactivity.

Previously we always had at least one `checkout` on the same node that the `˙git fetch` was called on, so the `known_hosts` was pre-populated. But this PR gets rid of this, and only uses `unstash` before `fetch` - so if we get scheduled on a "fresh" node that had no `checkout` steps executed on it since it spun up, we get an error.

bensze01: I've filed the issue:
https://linaro.atlassian.net/browse/TFC-747

bensze01: @ronald-cron-arm I would actually prefer that we hardcode `github.com` into the `ssh-keyscan` workaround instead of trying to parse the origin url, since we don't use any other hosts, and we expect that this will be short-term workaround until Linaro fix this on their side.

ronald-cron-arm:
I've rebased following the merge of #203 and removed the commit about the GitHub URL. The potential issue with `gen_abi_api_checking_job` and parametrized jobs will be addressed through  https://linaro.atlassian.net/browse/TFC-747.

vars/gen_jobs.groovy
ronald-cron-arm: I've removed that change.

gilles-peskine-arm (approved):
LGTM at b4b957ede246ee6f2fea0749709a84a2985a48fa on code reading.

bensze01 (approved):
Looks good to me, pending the CI results.

ronald-cron-arm:
All tests have run successfully. I am merging it then.

Issue #202, opened 2025-04-23
Switch MSVC builds to Visual Studio 2017


gilles-peskine-arm:
We currently test all branches with Visual Studio 2013 in PR jobs, and with both 2013 and 2017 in release jobs. Mbed TLS 3.6 LTS [officially requires Microsoft Visual Studio 2017 or later](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.6.0/README.md#tool-versions). Now that Mbed TLS 2.28 LTS is out of support, switch to 2017 everywhere.

The CI scripts support complicated logic to select VS versions. Cleaning up this logic is out of scope, and in fact probably undesirable since it's likely that we'll want to test TF-PSA-Crypto 1.x and Mbed TLS 4.x with newer VS versions in the future.

Issue #27, opened 2021-12-16
Unify Windows testing between PR and release jobs


gilles-peskine-arm:
The nightly/release jobs run more Windows builds than the PR job: the PR job only runs `windows_testing.py` with Visual Studio 2013 and mingw, whereas the nightly/release jobs test with all supported versions of Visual Studio (in 2.2x and 3.x, at the time of writing, that's 2015 and 2017). This is rarely a problem in practice because we rarely change Windows-specific things and we test with the oldest supported version, but it does introduce risk when merging a PR.

The goal of this task is to make Windows testing identical in all CI jobs, without meaningfully reducing the test coverage and without significantly increasing the typical lead time for a run of pr-head.

To keep the load moderate, we may want to do https://github.com/ARMmbed/mbedtls-test/issues/26 first.

Also, to keep the lead time for pr-head low, we may want to run some mostly-but-not-completely redundant tests only in pr-merge (and in non-PR jobs) and not in pr-head.

gilles-peskine-arm:
https://github.com/Mbed-TLS/mbedtls/pull/7462 passed VS2013 (the pull request CI passed) but [broke VS2015](https://github.com/Mbed-TLS/mbedtls/pull/7502#issuecomment-1524858851). This is the first time I remember seeing this happen.

PR #203, opened 2025-04-23
Switch MSVC builds to Visual Studio 2017

gilles-peskine-arm:
Resolves https://github.com/Mbed-TLS/mbedtls-test/issues/202. Also resolves https://github.com/Mbed-TLS/mbedtls-test/issues/27.

Scope: I've deliberately made a minimal set of changes. There are probably many simplifications we can make, but that was already the case before (we stil have code related to branches older than 2.18 when we didn't require a C99 compiler!). The current setup is simpler than it might become in the future, however, because currently it's fine to test all branches against the same VS version, but there will probably be times when we want different versions in LTS branches.

Test jobs:
* [release job, mbedtls/development, OpenCI](https://mbedtls.trustedfirmware.org/job/mbedtls-release-ci-testing/358/) → looks ok to me (passed and seems to have used the intended VS version throughout)
* [release job, mbedtls/development, internal CI](https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/798/) → looks ok to me (passed and seems to have used the intended VS version throughout)

### DIFF
--- a/vars/common.groovy
+++ b/vars/common.groovy
@@ -376,11 +376,9 @@ void check_every_all_sh_component_will_be_run(Collection<BranchInfo> infos) {
 
 def get_supported_windows_builds() {
     def vs_builds = []
-    if (env.JOB_TYPE == 'PR') {
-        vs_builds = ['2013']
-    } else {
-        vs_builds = ['2013', '2017']
-    }
+    /* At the time of writing, all supported branches (3.6, development)
+     * advertise support for Visual Studio 2017 (VS 15.0) and above. */
+    vs_builds = ['2017']
     echo "vs_builds = ${vs_builds}"
     return ['mingw'] + vs_builds
 }
--- a/vars/scripts.groovy
+++ b/vars/scripts.groovy
@@ -37,20 +37,20 @@ cmake -D CMAKE_BUILD_TYPE:String=Check -G "MinGW Makefiles" . || exit
 mingw32-make lib || exit
 '''
 
-@Field static final String win32_msvc12_32_test_bat = '''\
-call "C:\\Program Files (x86)\\Microsoft Visual Studio 12.0\\VC\\vcvarsall.bat" || exit
+@Field static final String win32_msvc15_32_test_bat = '''\
+call "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Community\\VC\\Auxiliary\\Build\\vcvarsall.bat" || exit
 set CC=cl
 if exist scripts\\make_generated_files.bat call scripts\\make_generated_files.bat || exit
-cmake . -G "Visual Studio 12" || exit
+cmake . -G "Visual Studio 15 2017" -A Win32 || exit
 MSBuild ALL_BUILD.vcxproj || exit
 programs\\test\\Debug\\selftest.exe || exit
 '''
 
-@Field static final String win32_msvc12_64_test_bat = '''\
-call "C:\\Program Files (x86)\\Microsoft Visual Studio 12.0\\VC\\vcvarsall.bat" || exit
+@Field static final String win32_msvc15_64_test_bat = '''\
+call "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Community\\VC\\Auxiliary\\Build\\vcvarsall.bat" || exit
 set CC=cl
 if exist scripts\\make_generated_files.bat call scripts\\make_generated_files.bat || exit
-cmake . -G "Visual Studio 12 Win64" || exit
+cmake . -G "Visual Studio 15 2017" -A x64 || exit
 MSBuild ALL_BUILD.vcxproj || exit
 programs\\test\\Debug\\selftest.exe || exit
 '''
--- a/vars/scripts.groovy
+++ b/vars/scripts.groovy
@@ -38,7 +38,7 @@ mingw32-make lib || exit
 '''
 
 @Field static final String win32_msvc15_32_test_bat = '''\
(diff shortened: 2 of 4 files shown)


vars/common.groovy
```diff
@@ -376,11 +376,9 @@ void check_every_all_sh_component_will_be_run(Collection<BranchInfo> infos) {
...
 def get_supported_windows_builds() {
     def vs_builds = []
-    if (env.JOB_TYPE == 'PR') {
-        vs_builds = ['2013']
-    } else {
-        vs_builds = ['2013', '2017']
-    }
+    /* At the time of writing, all supported branches (3.6, development)
+     * advertise support for Visual Studio 2017 (VS 15.0) and above. */
+    vs_builds = ['2017']
```
bjwtaylor, line 381: Not a topic for discussion on this PR, however we might want to think about the phrasing of 2017 and above? If we only test 2017, we presumably can only reliably support 2017 and later versions are an unknown quantity and no guarantees can be provided?

bjwtaylor (approved):

vars/common.groovy
gilles-peskine-arm: We can't test with everything... Generally, with “positive” tooling (things that build), if an old version works, we conclude that it has enough features and we assume that more recent versions also have those features. With “negative” tooling (things that check), if a recent version works, we conclude that our product is good and that older versions of the tool would also pass it. So when we test with an old compiler, we conclude that we don't rely on parts of the language that this compiler doesn't implement. But we can't conclude that our code would compile without warnings with a more recent compiler.

Regarding Visual Studio specifically, to test with more recent versions, we'd have to change our Windows images, which is a pain that we've been putting off. We can put it off as long as our code remains compatible with VS 2017.

vars/scripts.groovy
```diff
@@ -37,20 +37,20 @@ cmake -D CMAKE_BUILD_TYPE:String=Check -G "MinGW Makefiles" . || exit
...
 '''
 
-@Field static final String win32_msvc12_32_test_bat = '''\
-call "C:\\Program Files (x86)\\Microsoft Visual Studio 12.0\\VC\\vcvarsall.bat" || exit
+@Field static final String win32_msvc15_32_test_bat = '''\
+call "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Community\\VC\\Auxiliary\\Build\\vcvarsall.bat" x86 || exit
 set CC=cl
 if exist scripts\\make_generated_files.bat call scripts\\make_generated_files.bat || exit
-cmake . -G "Visual Studio 12" || exit
+cmake . -G "Visual Studio 15 2017" -A Win32 || exit
```
gilles-peskine-arm, line 44: This change is fine [on OpenCI](https://mbedtls.trustedfirmware.org/job/mbedtls-release-ci-testing/357/execution/node/705/log/?consoleFull), but [breaks internal CI](https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-ci-testing/796/execution/node/613/log/) and I don't understand why. On the internal CI, CMake is unhappy:

```
C:\builds\workspace\mbedtls-release-ci-testing\src>call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat" x86   || exit
**********************************************************************
** Visual Studio 2017 Developer Command Prompt v15.0.26730.16
** Copyright (c) 2017 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x86'
CMake Error: The source directory "C:/Users/Administrator/source" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
```

In contrast, CMake proceeds normally on OpenCI:
```
 c:\windows\workspace\mbedtls-release-ci-testing\src>call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat" x86   || exit
 **********************************************************************
 ** Visual Studio 2017 Developer Command Prompt v15.9.40
 ** Copyright (c) 2017 Microsoft Corporation
 **********************************************************************
 [vcvarsall.bat] Environment initialized for: 'x86'
 -- Selecting Windows SDK version 10.0.17763.0 to target Windows 10.0.14393.
 -- The C compiler identification is MSVC 19.16.27045.0
…
```

I see that the version of Visual Studio is different, but I don't see how this would prevent CMake from finding `CMakeLists.txt`. It's possible that the version of CMake is different, but again I don't see how that would be a problem: we're only making basic use of CMake.

gilles-peskine-arm: Maybe CMake on internal CI is antique? The `-A` option was added [in CMake 3.1.0 released in Dec 2014](https://cmake.org/pipermail/cmake/2014-December/059418.html).

gilles-peskine-arm: Actually, what's with `C:/Users/Administrator/source`? Looks like the slightly older VS 2017 release on the internal CI changes the current directory when `vcvarsall.bat` is called!

minosgalanakis (approved):
LGTM

vars/common.groovy
irwir: > we might want to think about the phrasing of 2017 and above?

Visual Studio can also apply different platform toolsets (for example, v141_xp instead of the default v143 for VS 2022). With run time well over 3 hours per commit, exaustive testing might become impractical in CI.
Testing of newer VS versions might be contemplated as a one-time final step before a release.

gilles-peskine-arm:
This also happens to resolve https://github.com/Mbed-TLS/mbedtls-test/issues/27 (thanks Bence) — the choice of VS versions was the only difference (wrt Windows) between the PR and release jobs.

bjwtaylor (approved):

Issue #189, opened 2024-12-12
Allow the restricted CI to fetch commits from restricted submodules


bensze01:
PRs in the restricted repo need to be able to refer to submodule commits that don't exist in the pubic submodule repos, only in the restricted ones.

gilles-peskine-arm:
A simple way to run the mbedtls CI with restricted submodule branches is to change the repository URLs in `.gitmodules` to the `-restricted` variants. This currently works: all it requires is that the CI can access the restricted module.

However, doing this drives us into a trap when it comes to make a release. We must not release a branch where `.gitmodules` points to a restricted repository. So if we've edited `.gitmodules`, we need to revert that change _and then not run the CI_.

A full resolution of this issue needs to ensure that we have a working release process. Update the release process (probably [shift 2](https://confluence.arm.com/display/CESW/Shift+2) (private link)) as necessary.

ronald-cron-arm:
Some thoughts about that. If we change the URL in ".gitmodules" to test the restricted PRs then some commits in restricted repositories will have ".gitmodules" pointing to restricted repositories. When these commits become public through a release, a user with no access to the restricted repositories cannot checkout these commits. Because of this and the extra work of fiddling  with `.gitmodules` during the release process as pointed by @gilles-peskine-arm, I am thinking that we should rather never change ".gitmodules" and adapt the CI to switch to the restricted URLs when testing a restricted PR. Wdyt?

ronald-cron-arm:
"When these commits become public through a release, a user with no access to the restricted repositories cannot checkout these commits"
That's probably actually not true. As after a release, all previously restricted commits, that are now public, are fetched as part of the fetch of the released branch and its git history. Thus for example, when checking out an Mbed TLS commit with ".gitmodules" pointing to the restricted repos and pointing to a now public TF-PSA-Crypto commit, `git submodule update` should just work as the TF-PSA-Crypto is already in the local git history.

ronald-cron-arm:
It seems that when you run `git submodule update tf-psa-crypto' and the TF-PSA-Crypto commit pointed to by the current Mbed TLS commit has already been fetched (by a previous `git submodule update`) then there is no remote access to fetch it again that could use the URL in the `.gitmodules` of the current Mbed TLS commit.
[...]

PR #206, opened 2025-05-14
Test restricted PRs with restricted submodules

bensze01:


vars/checkout_repo.groovy
```diff
@@ -75,33 +75,42 @@ void checkout_framework_repo(BranchInfo info) {
...
 
 void checkout_tf_psa_crypto_repo(BranchInfo info) {
-    if (env.TARGET_REPO == 'tf-psa-crypto' && env.CHECKOUT_METHOD == 'scm') {
-        checkout_report_errors(scm)
-        if (!info.framework_override) {
-            if (!isUnix()) {
-                throw new IllegalStateException("The first checkout of the framework must be made on a Unix node")
+    try {
+        sh '''
+git config --global url.git@github.com:Mbed-TLS/mbedtls-framework-restricted.insteadof https://github.com/Mbed-TLS/mbedtls-framework
```
ronald-cron-arm, line 80: To me, we should not do that in the case of the testing of a public TF-PSA-Crypto PR, right?
Otherwise what about to have all these aliases once for all in `checkout_repo`?

bensze01: You are right, I meant to make these conditional on `env.IS_RESTRICTED`.
When I wrote this I was thinking seting them for the whole checkout process might mess something up - but I can't think of any scenario where that would apply at the moment. So I'll move them.

ronald-cron-arm (changes requested):

vars/environ.groovy
```diff
@@ -34,19 +34,40 @@ void set_pr_environment(String target_repo, boolean is_production) {
...
         switch (target_repo) {
             case 'framework': //fallthrough
             case 'tf-psa-crypto':
-                env.FRAMEWORK_REPO = 'git@github.com:Mbed-TLS/mbedtls-framework.git'
-                env.TF_PSA_CRYPTO_REPO = 'git@github.com:Mbed-TLS/TF-PSA-Crypto.git'
-                env.MBED_TLS_REPO = 'git@github.com:Mbed-TLS/mbedtls.git'
+                if (env.IS_RESTRICTED) {
+                    env.FRAMEWORK_REPO = 'git@github.com:Mbed-TLS/mbedtls-framework-restricted.git'
+                    env.TF_PSA_CRYPTO_REPO = 'git@github.com:Mbed-TLS/TF-PSA-Crypto-restricted.git'
+                    env.MBED_TLS_REPO = 'git@github.com:Mbed-TLS/mbedtls-restricted.git'
```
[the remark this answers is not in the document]
bensze01: This should be fixed by 3f6f7a0e348262c533b663a2cfd370fc4bb8751a

vars/checkout_repo.groovy
```diff
@@ -157,15 +143,39 @@ void checkout_repo(BranchInfo info) {
...
-                    switch (info.repo) {
-                        case 'tls':
-                            checkout_tls_repo(info)
-                            break
-                        case 'tf-psa-crypto':
-                            checkout_tf_psa_crypto_repo(info)
-                            break
-                        default:
-                            error("Invalid repo: ${info.repo}")
+                    try {
```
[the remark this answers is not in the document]
bensze01: I think this was caused by the outdated mbedtls-restricted test PR. I've pushed a new version that includes the tf-psa-crypto submodule.

gilles-peskine-arm:
Do you have test runs where the submodules are restricted? In the runs that are currently linked from the PR description, the commit under test is restricted, but the submodules are public.

bensze01:
@ronald-cron-arm I think I addressed all of your feedback
@gilles-peskine-arm I've added some test runs that you've requested for mbedtls-restricted, I'll add the ones for the rest of the shortly.

ronald-cron-arm:
> Do you have test runs where the submodules are restricted? In the runs that are currently linked from the PR description, the commit under test is restricted, but the submodules are public.

The current plan is to have ".gitmodules" to always point to the public repositories as discussed in https://github.com/Mbed-TLS/mbedtls-test/issues/189. Thus we do not really need test runs where the submodules are restricted. If we add some CI tests around that it would be to fail the CI if one of the URLs in ".gitmodules" is a restricted one.

ronald-cron-arm:
> @ronald-cron-arm I was thinking it might get annoying if eg. we are only working on restricted changes in Mbed TLS, having to keep both restricted submodule repositories updated. Since the change required was fairly small for the workaround, I'd prefer to keep it, but I can remove it if you are more comfortable with that.

I'd prefer to remove it. Two arguments:
1) We do not need it as we can update restricted repositories if needed. Even if the change is quite small, it complicates the code and its understanding.
2) We should rather promote the regular update of the restricted repositories with the work on public branches. That way we minimize the risk of issues when doing releases.

gilles-peskine-arm:
I'm weakly inclined to include the fallback to the non-restricted repository. Having to ensure that either all (sub)modules are public or all are restricted adds complication and surprise when we push something for testing. Not just pull requests, but also just doing ad hoc testing, e.g. doing investigation into a potential security hole. I'd rather have a little extra complexity in the code (it's really not much) than a predictable recurring theme of “oops, I forgot to push one more sha to one more restricted repo and now I have to trigger a new job and wait again”.

gilles-peskine-arm:
> In that context of restricted pull requests, I do not think we should push some well chosen sha to restricted repos. If we need a more recent version of a submodule that is currently in the public repo but not in the restricted one, we need to update the restricted one with a merge of the public development branch into the development-restricted one.

When we merge something, yes. But not when we're just testing, perhaps to investigate a vulnerability. At that stage it there's no reason to touch official branches, so it would really be pushing ad hoc shas.

ronald-cron-arm:
> > In that context of restricted pull requests, I do not think we should push some well chosen sha to restricted repos. If we need a more recent version of a submodule that is currently in the public repo but not in the restricted one, we need to update the restricted one with a merge of the public development branch into the development-restricted one.
> 
> When we merge something, yes. But not when we're just testing, perhaps to investigate a vulnerability. At that stage it there's no reason to touch official branches, so it would really be pushing ad hoc shas.

Thus you are referring  to a "draft test PR" in the restricted repo?

For that kind of use case I would rather use the `mbed-tls-restricted-pr-test-parametrized` job. It has all the flexibility we need and we do not have to create a "draft test PR", we can just test a branch on our fork.

gilles-peskine-arm:
> I'm weakly inclined to include the fallback to the non-restricted repository. Having to ensure that either all (sub)modules are public or all are restricted adds complication and surprise when we push something for testing.

Case in point: I just naturally made a restricted branch with a public framework in https://github.com/Mbed-TLS/mbedtls-restricted/pull/1362. Granted, right now I had to — but it's what I would have done even after restricted submodules are supported on the CI.

So now I'm strongly inclined to include the fallback.

gilles-peskine-arm:
> Otherwise have you planned to do a full review of this PR?

No, as discussed on Slack with Bence, given my current review bandwidth, I'd rather not.

ronald-cron-arm:
> > Otherwise have you planned to do a full review of this PR?
> 
> No, as discussed on Slack with Bence, given my current review bandwidth, I'd rather not.

That's completely fine, no worries. @minosgalanakis can help there and has already assigned himself it seems.

vars/mbedtls.groovy
```diff
@@ -157,7 +157,14 @@ void run_job() {
 }
 
 void run_framework_pr_job() {
-    run_pr_job('framework', true, ['development', 'mbedtls-3.6'], ['development'])
+    environ.parse_scm_repo()
+    def tls_branches = ['development', 'mbedtls-3.6']
```
minosgalanakis, line 161: Why not use tenary assigment for tls_branchs and tf_psa_crypto_branches? 

     env.IS_RESTRICTED ? ['development-restricted', 'mbedtls-3.6-restricted']
                                       : ['development', 'mbedtls-3.6']
                                       
Update: We may need to set it as an intermediate variable since strings are treated as boolean true even if empty, so not sure if its worth the risk.

minosgalanakis (commented):
Looks good overall, minor comments from my side.

vars/mbedtls.groovy
bensze01: It's actually the opposite. Groovy strings are truthy if and only if  `string.length() > 0`.

vars/mbedtls.groovy
```diff
@@ -157,7 +157,14 @@ void run_job() {
...
+    def tf_psa_crypto_branches = ['development']
+    if (env.IS_RESTRICTED) {
+        tls_branches = ['development-restricted', 'mbedtls-3.6-restricted']
+        tf_psa_crypto_branches = ['development-restricted']
+    }
+    run_pr_job('framework', true, tls_branches, tf_psa_crypto_branches)
```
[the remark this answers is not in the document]
bensze01: It's not coupled - `production` is set for all PR jobs. It's meaning is effectively `! isParametrized`.

minosgalanakis (dismissed):
LGTM

ronald-cron-arm (approved):
LGTM, thanks.

bensze01:
I've pushed the OpenCI config change that depends on this PR to Gerrit: [mbed-tls-job-configs/+/39235](https://review.trustedfirmware.org/c/ci/mbedtls/mbed-tls-job-configs/+/39235)

bensze01:
All the CI runs passed

gilles-peskine-arm:
> > So now I'm strongly inclined to include the fallback.
>
> That's the plan now. Otherwise have you planned to do a full review of this PR?

I'm now confused. Is `.gitmodules` used or not?

Please reply by updating the [developer guide](https://github.com/Mbed-TLS/mbedtls-test/blob/main/developer_guide.md) and the [framework guide](https://github.com/Mbed-TLS/mbedtls-framework/blob/main/docs/framework-design.md#ci-architecture) (probably with the information in one document, and a pointer in the other document).

PR #212, opened 2025-07-15
Disable ipv6 networking inside docker

bensze01:
Workaround for https://github.com/Mbed-TLS/mbedtls-test/issues/213

Combined test runs with Mbed-TLS/mbedtls#10302:
- [Open CI](https://mbedtls.trustedfirmware.org/job/mbed-tls-restricted-pr-test-parametrized/164/)
- [Internal CI](https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-test-parametrized/516/)

PR #216, opened 2025-07-17
Fix priority inversion in init-docker-images stage

bensze01:
Taking the lock before we are running on a node is preventing the priority sorter from properly prioritizing merge-queue jobs.
Invert the order of taking the "locks" (node -> docker-image), to fix this priority inversion.

Test runs (#216 cherry-picked on top of #212) :
- development
  - [Internal CI][1]
  - [Open CI][2]

**PLEASE NOTE**: Care should be taken when merging this PR, since the interaction between runs that still use the old CI scripts (eg. started before the merge) and new runs could cause a deadlock. It would be best to merge it in a quiet period on the CI.

[1]: https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-test-parametrized/518/
[2]: https://mbedtls.trustedfirmware.org/job/mbed-tls-restricted-pr-test-parametrized/164/

### DIFF
--- a/vars/gen_jobs.groovy
+++ b/vars/gen_jobs.groovy
@@ -708,10 +708,13 @@ def gen_dockerfile_builder_job(String platform, boolean overwrite=false) {
     common.docker_tags[platform] = tag
 
     return job(platform) {
-        /* Take the lock on the master node, so we don't tie up an executor while waiting */
-        lock(tag) {
-            def node_label = arch == 'amd64' ? 'dockerfile-builder' : "container-host-$arch"
-            analysis.node_record_timestamps(node_label, platform) {
+        def node_label = arch == 'amd64' ? 'dockerfile-builder' : "container-host-$arch"
+        analysis.node_record_timestamps(node_label, platform) {
+            /* Take the lock only once we are running on a node.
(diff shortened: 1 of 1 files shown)


vars/gen_jobs.groovy
```diff
@@ -708,10 +708,13 @@ def gen_dockerfile_builder_job(String platform, boolean overwrite=false) {
     common.docker_tags[platform] = tag
 
     return job(platform) {
-        /* Take the lock on the master node, so we don't tie up an executor while waiting */
```
gilles-peskine-arm, line 711: I guess it's not so important that an executor gets tied up while waiting for the lock, because we don't expect much contention? Since the critical section will almost always just see that the image is there and exit quickly, and if an image is getting rebuilt then we don't really need the nodes for anything until that's done anyway?

gilles-peskine-arm (approved):
I can't fully appreciate which order is best, but your reasoning seems plausible to me. Approved subject to seeing a few successful CI runs.

vars/gen_jobs.groovy
bensze01: It also might increase our AWS bill by a few cents, but we are billed by CPU time, so idling executors shouldn't cost that much.

gilles-peskine-arm (approved):
LGTM

Issue #235, opened 2025-12-04
Result analysis is not visible in the Pipeline Overview on Jenkins


gilles-peskine-arm:
In the “Pipeline Overview” a.k.a. “Graph” view (`…/pipeline-graph/`) on Jenkins, which is the default view on the new OpenCI, the step of result analysis that does the actual work (tls-result-analysis or tf-psa-crypto-result-analysis) is not shown. This is especially confusing when that step fails, since there's no visible indication of what fails.

For example, [tls-result analysis failed but isn't visible in the graph view](https://ci.trustedfirmware.org/job/mbed-tls-tf-psa-crypto-multibranch/job/PR-575-head/13/pipeline-graph/):
<img width="311" height="156" alt="Image" src="https://github.com/user-attachments/assets/f6444be1-dc53-4899-8d57-3c5ee64ea378" />

The [“Blue Ocean” view](https://ci.trustedfirmware.org/blue/organizations/jenkins/mbed-tls-tf-psa-crypto-multibranch/detail/PR-575-head/13/pipeline/) (`/blue/…`) shows this correctly:
<img width="441" height="149" alt="Image" src="https://github.com/user-attachments/assets/954d1393-64b8-48b5-b1a1-6018a850f49b" />

You can also find the information on the [“Pipeline Steps” view](https://ci.trustedfirmware.org/job/mbed-tls-tf-psa-crypto-multibranch/job/PR-575-head/13/flowGraphTable/) if you browse the very long list. (Tip: install the userscript mentioned in the [CI failure FAQ](https://mbed-tls.readthedocs.io/en/latest/kb/testing/understanding-ci-failures/#pr-tests-failed-which-components-failed)).

I don't know if this is a bug in our Groovy code or something that needs to be fixed in Jenkins or a Jenkins plugin.

PR #236, opened 2025-12-08
Work around bug in graph-view display

bensze01:
Fixes: #235 

The bug on the new graph view page described in #235 seems to be triggered by having nested nodes (calls to `stage` or `parallel`) with identical names.

Eg. the following will tigger the bug, as it does in `gen_docker_job()`:

```groovy
parallel foo: {
    stage('foo') {
        sh 'echo bar'
    }
}
```

This PR works around the bug by changing the name of the second inner stage in `gen_docker_job()`

CI run
- https://ci.trustedfirmware.org/job/mbed-tls-restricted-pr-test-parametrized/37/pipeline-graph/
  -  (the failure is caused by the example Crypto PR used to demonstrate the issue in the analysis step, its not related to this PR)

### DIFF
--- a/vars/gen_jobs.groovy
+++ b/vars/gen_jobs.groovy
@@ -134,7 +134,7 @@ Map<String, Callable<Void>> gen_docker_job(Map<String, Closure> hooks,
                     }
                 }
             }
-            stage(job_name) {
+            stage("${job_name}.inner") {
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm (approved):
LGTM

Thanks for looking into this!

PR #237, opened 2025-12-08
Clone all submodules without groovy handlers

bensze01:
List and clone all submodules that are still unitialized after all the handlers have run.

Required by Gilles's PR adding mldsa-native as a submodule.

CI runs:
- @https://github.com/Mbed-TLS/mbedtls-test/commit/dfc2953d3f8353038456d6b13b42ce97da4d3869
  - Mbed-TLS/TF-PSA-Crypto#593 @[6aef476](https://github.com/gilles-peskine-arm/TF-PSA-Crypto/commit/6aef476a1197d7d603cf29b652b44a769e5cb23e)
    - [mbed-tls-restricted-pr-test-parametrized][a2] 
- @https://github.com/Mbed-TLS/mbedtls-test/commit/f32f0d37443a82224c23bd1aa75aa3fb7724050c
  - Mbed-TLS/TF-PSA-Crypto#593 @[6aef476](https://github.com/gilles-peskine-arm/TF-PSA-Crypto/commit/6aef476a1197d7d603cf29b652b44a769e5cb23e)
    - [mbed-tls-restricted-pr-test-parametrized][a1] → OK did check out the submodule (the tf-psa-crypto builds in the full config passed, like in https://ci.trustedfirmware.org/job/mbed-tls-tf-psa-crypto-multibranch/job/PR-593-head/2/ with a workaround in `all-core.sh`, and unlike in https://ci.trustedfirmware.org/job/mbed-tls-tf-psa-crypto-multibranch/job/PR-593-head/1/ with no workaround)
  - TF-PSA-Crypto development
    - [mbed-tls-restricted-pr-test-parametrized][b1] → OK (one component failed due to a network glitch, but otherwise ran fine)
  - development
    - [mbed-tls-restricted-pr-test-parametrized][c1] → PASS
  - mbedtls-3.6
    - [mbed-tls-restricted-pr-test-parametrized][d1] → PASS
  
[a1]: https://ci.trustedfirmware.org/view/Mbed-TLS/job/mbed-tls-restricted-pr-test-parametrized/33/
[b1]: https://ci.trustedfirmware.org/view/Mbed-TLS/job/mbed-tls-restricted-pr-test-parametrized/34/
[c1]: https://ci.trustedfirmware.org/view/Mbed-TLS/job/mbed-tls-restricted-pr-test-parametrized/35/
[d1]: https://ci.trustedfirmware.org/view/Mbed-TLS/job/mbed-tls-restricted-pr-test-parametrized/36/
[a2]: https://ci.trustedfirmware.org/view/Mbed-TLS/job/mbed-tls-restricted-pr-test-parametrized/54/

### DIFF
--- a/vars/checkout_repo.groovy
+++ b/vars/checkout_repo.groovy
@@ -194,9 +194,6 @@ void checkout_repo(BranchInfo info) {
 git submodule init
 git submodule foreach 'git submodule init'
 
-# Only split unquoted expansions on newlines
-IFS='
-'
 # Get list of missing submodules (lines starting with '-')
 missing_submodules=$(git submodule status --recursive | awk '/^-/{print $2}')
 for module in $missing_submodules; do


vars/checkout_repo.groovy
```diff
@@ -188,6 +188,20 @@ void checkout_repo(BranchInfo info) {
...
+# Initialize submodule variables (required by `git submodule status`)
+git submodule init
+git submodule foreach 'git submodule init'
+
+# Only split unquoted expansions on newlines
+IFS='
+'
+for module in $(git submodule status --recursive | sed -n 's/^-[^ ]* //p'); do
+    git -C "$module" submodule update --recursive --init --depth=1 .
+done
```
ronald-cron-arm: This seems correct but it is quite hard to follow, especially "| sed -n 's/^-[^ ]* //p'". Please rewrite it with comments and  probably an intermediate `uninitialized_modules` variable:
```
# Get a list of uninitialized submodules (lines starting with '-')
uninitialized_modules=$(git submodule status --recursive | awk '/^-/{print $2}')
```

gilles-peskine-arm (dismissed):
LGTM

vars/checkout_repo.groovy
```diff
@@ -197,7 +197,9 @@ git submodule foreach 'git submodule init'
 # Only split unquoted expansions on newlines
 IFS='
 '
-for module in $(git submodule status --recursive | sed -n 's/^-[^ ]* //p'); do
+# Get list of missing submodules (lines starting with '-')
+missing_submodules=$(git submodule status --recursive | awk '/^-/{print $2}')
+for module in $missing_submodules; do
     git -C "$module" submodule update --recursive --init --depth=1 .
```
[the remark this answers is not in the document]
bensze01: In case one of the submodules has its own submodule.
Eg. if mldsa-native ever adds a submodule of its own.

vars/checkout_repo.groovy
```diff
@@ -188,6 +188,22 @@ void checkout_repo(BranchInfo info) {
...
+git submodule foreach 'git submodule init'
+
+# Only split unquoted expansions on newlines
+IFS='
+'
+# Get list of missing submodules (lines starting with '-')
+missing_submodules=$(git submodule status --recursive | awk '/^-/{print $2}')
+for module in $missing_submodules; do
+    git -C "$module" submodule update --recursive --init --depth=1 .
+done
```
ronald-cron-arm: Like this the purpose of "IFS=" is not clear to me and it seems it would not handle well a submodule name with spaces.

What about something like (not tested):
```suggestion
# Get list of missing submodules (lines starting with '-')
missing_submodules=$(git submodule status --recursive | awk '/^-/{print $2}')

# Loop through each missing submodule (use read safely to handle whitespace in paths)
while IFS= read -r module; do
    git -C "$module" submodule update --recursive --init --depth=1 .
done <<< "$missing_submodules"
```

bensze01: Git command outputs generally escape all special characters in paths *except* space. This is documented in various commands,  though IIRC not `submodule status` specifically.

bensze01: The difference between the two is that we run `tests/scripts/all.sh` inside the docker container on linux - but this is not really a meaningful distinction at the moment, since all the other platforms run outside a container, and all platforms have access to the network.

I think network isolation inside the container could enable us to run a (subset of?) the containerized linux tests automatically without intervention for outside contributors, which would make the difference more meaningful - but considering the current load on our CI maybe that's not the greatest idea either way.

The python requirements would also need some special handling (eg. use a trusted version of `min_requirements.py`, download packages with network access enabled, then install them without network access), since both `min_requirements.py` and python packages built from source can run arbitrary commands.

ronald-cron-arm (approved):
LGTM, thanks.

PR #257, opened 2026-06-20
Throttle large parallel fan-out with throttle categoy

saheerb:
Wrap the main tls-testing parallel fan-out in the
mbetls-parallel-limit throttle category.

This was the recommended approach by cloudbees to reduce queue pressure.

### DIFF
--- a/vars/mbedtls.groovy
+++ b/vars/mbedtls.groovy
@@ -42,7 +42,9 @@ void run_tls_tests(Collection<BranchInfo> infos) {
 
         jobs.failFast = false
         analysis.record_inner_timestamps('main', 'run_pr_job') {
-            parallel jobs
+            throttle(['mbedtls-parallel-executor-limit']) {
+                parallel jobs
+            }
         }
     } catch (err) {
         def failed_names = infos.collectMany({ info -> info.failed_builds}).sort().join(" ")
@@ -217,7 +219,9 @@ void run_release_job(Collection<String> tls_branches, Collection<String> tf_psa_
                     jobs = common.wrap_report_errors(jobs)
                     jobs.failFast = false
                     analysis.record_inner_timestamps('main', 'run_release_job') {
-                        parallel jobs
+                        throttle(['mbedtls-parallel-executor-limit']) {
+                            parallel jobs
+                        }
                     }
                 }
             }


gilles-peskine-arm (dismissed):
LGTM on code reading. But I'm not a Jenkins or Groovy experts.

Have you run a test build? I'd like to see a passing build, and a build where some jobs fail.

saheerb:
> LGTM on code reading. But I'm not a Jenkins or Groovy experts.
> 
> Have you run a test build? I'd like to see a passing build, and a build where some jobs fail.

I changed the approach after recommendation from cloudbees. Test run: 

https://ci.test.trustedfirmware.org/view/Mbed-TLS/job/mbed-tls-framework-multibranch/job/PR-309-head/8/cloudbees-pipeline-explorer/

vars/mbedtls.groovy
```diff
@@ -42,7 +42,9 @@ void run_tls_tests(Collection<BranchInfo> infos) {
 
         jobs.failFast = false
         analysis.record_inner_timestamps('main', 'run_pr_job') {
-            parallel jobs
+            throttle(['mbetls-parallel-limit']) {
```
ronald-cron-arm:
```suggestion
            throttle(['mbedtls-parallel-limit']) {
```
? But maybe there is a typo somewhere else?

What is the maximum number of concurrent jobs?
Is it set to the number of executors?
I can see in https://ci.test.trustedfirmware.org/view/Mbed-TLS/job/mbed-tls-framework-multibranch/job/PR-309-head/8/cloudbees-pipeline-explorer/ that the testing jobs have started after about 5 minutes (compared to a few hours without the change), how do we explain that?

```
04:05:27 AM Still waiting to schedule task
04:05:27 AM All nodes of label ‘mbedtls-container-host’ are offline
04:06:43 AM Running on EC2 (ec2-mbedtls-jenkins-agents) - mbed-tls Linux AMI (i-0dbf3bd1daa77005b) in /home/admin/workspace/ramework-multibranch_PR-309-head
04:11:30 AM + aws ecr get-login-password --region eu-west-1
04:11:30 AM + docker login --username AWS --password-stdin 637423617205.dkr.ecr.eu-west-1.amazonaws.com
```

saheerb: >? But maybe there is a typo somewhere else?

Thanks, yes, that is a typo. It should be mbedtls-parallel-limit; I also had the same typo in the Jenkins throttle category (jenkins system variable), so I’ll fix both.


> What is the maximum number of concurrent jobs?

The current cap is set to 100 total concurrent builds for that category, meaning at most 100 throttled branches can be active/buildable at once across that controller for jobs using this category; it is not tied directly to the total number of executors. We will need to tune this a bit to find the best value. In practice, this cap becomes the effective upper bound on concurrency for these throttled Pipeline branches, regardless of how many executors may be available on a particular label.

>hat the testing jobs have started after about 5 minutes (compared to a few hours without the change), how do we explain that?

As per CloudBees, the main problem was that letting ~1000 parallel node { ... } branches enter the Jenkins queue at once made queue maintenance expensive and held the global queue lock long enough to delay executor assignment, agent lifecycle work, and branches entering the node body.
 What we seem to be seeing now is that the first few minutes are mostly the initial Pipeline fan-out phase where Jenkins creates and manages the queued parallel work, but after that executors are picked up much more quickly. The issue before was that with ~1000 items in the queue, that handoff could take hours before branches actually started running on executors. I still don’t know what changed recently to make the controller much more sensitive to this pattern.

gilles-peskine-arm (dismissed):
Not sure how it helps, but ok on code reading. Approved pending validation through a CI run.

Can you please run a test job on the normal CI instance (not the test instance, which might be configured differently), e.g. the PR job on https://github.com/Mbed-TLS/mbedtls-restricted/pull/906 ?

saheerb:
> Not sure how it helps, but ok on code reading. Approved pending validation through a CI run.
> 
> Can you please run a test job on the normal CI instance (not the test instance, which might be configured differently), e.g. the PR job on https://github.com/Mbed-TLS/mbedtls-restricted/pull/906 ?

I dont think I can test this change in PROD without merging it in. At least not without editing job/pipeline lib config. What I suggest is to bring this in and if it doesn't work we can alway revert.

ronald-cron-arm:
Currently the CI is working as expected and we have a fair amount of PR under test. I would not merge that change right now.

PR #259, opened 2026-08-06
Mount /opt/host from the Docker host

gilles-peskine-arm:
1. Clean up some code related to obsolete CI instances.
2. Update the mbedtls-dockerfile-builder job to match the current job configuration.
3. Install some cross-compilers from Ubuntu.
4. In the modern Linux Dockerfile and in the arm-compilers, mount /opt/host from the host. The short-term objective is to have Qemu.

I'm only touching the most recent Ubuntu image that we have, because we should try to wind down the old ones. That image is 24.04, and we should move to 26.04 one of these days, but adding a new image is out of scope here.

Companion PR to add Qemu to the Jenkins executor where docker runs: https://review.trustedfirmware.org/c/ci/aws-amis/+/53424

Test runs on 541ae5912a2bc8d9f063c667a5c919fe7533ac52 (identical content to 913e5939c581805eb808e1b7df079de07ca16083):

* Dockerfile builder: https://ci.trustedfirmware.org/view/Mbed-TLS/job/mbedtls-dockerfile-builder/10/ → PASS
* With the old executor image, Mbed TLS development: https://ci.trustedfirmware.org/view/Mbed-TLS/job/mbedtls-restricted-release-ci-testing/31/ → PASS
* With the new executor image, Mbed TLS development: TODO


bensze01 (dismissed):
LGTM!

resources/docker_files/run.sh
```diff
@@ -52,5 +52,5 @@ echo "  User ID:Group ID --> $USR_ID:$USR_GRP"
 echo "  Mounting $MOUNT_DIR --> /var/lib/ws"
 echo "****************************************************"
 
-sudo docker run --network=host --rm -i -t -u $USR_ID:$USR_GRP -w /var/lib/ws -e HOME=/var/lib/ws -v $MOUNT_DIR:/var/lib/ws --cap-add SYS_PTRACE ${IMAGE}
+sudo docker run --network=host --rm -i -t -u $USR_ID:$USR_GRP -w /var/lib/ws -e HOME=/var/lib/ws -v $MOUNT_DIR:/var/lib/ws -v /opt/host --cap-add SYS_PTRACE ${IMAGE}
```
bjwtaylor: Should this be -v /opt/host:/opt/host?

yiwu0b11 (commented):
(comment send in wrong place)

vars/common.groovy
```diff
@@ -233,7 +222,13 @@ String docker_script(
 ) {
     def docker_image = get_docker_tag(platform)
     def env_args = env_vars.collect({ e -> "-e $e" }).join(' ')
-    def volume_args = volumes.collect({ v -> "-v $v" }).join(' ')
+
+    def volume_list = volumes.toList()
+    if (!volume_list.any({ v -> v == '/opt/host' || v.endsWith(':/opt/host') })) {
```
yiwu0b11: *should this be read-only? like
```
volume_list.add('/opt/host:/opt/host:ro')
```

gilles-peskine-arm: Good point. Indeed, the container shouldn't be writing to the host.

bjwtaylor (approved):

yiwu0b11 (approved):

